Issue #8, opened 2023-03-09
Make setup does not work on MacOS only on Linux based machines


shiroyasha:
Found the first bug!

`make setup` does not work on MacOS stating that it can't set the UID parameter.
Screenshot:

![Screenshot 2023-03-09 at 19 29 39](https://user-images.githubusercontent.com/1779493/224121905-defc5447-3e19-417b-b5fc-6ba4af89d2aa.png)

benj0c:
> It's a good idea to use lowercase variables in your shell scripts, as uppercase variables are usually reserved for shell internals. (Environment variables are an exception to this, but should really be treated the same way, as things with special meanings that you shouldn't arbitrarily change without understanding what you're doing.) In this case, bash sets $UID to the Unix uid it's running under, and won't allow it to be changed or the variable to be used for anything else. Shells other than bash (sh may be dash or some other shell on some platforms) may not use $UID that way, and if bash was built to work in Bourne shell compatibility mode when invoked as sh (which is the default, but many Linux distributions disable that) it won't treat $UID as special either.

Resource: https://serverfault.com/questions/259843/bash-script-error-test-line-5-uid-readonly-variable

benj0c:
Once this issue with passing the `UID` and `GID` args is solved, we'll face another problem when group id already exists on guest machine:

```
#0 0.163 groupadd: GID '20' already exists
```

shiroyasha:
Closing. Fixed with PR #9.

PR #9, opened 2023-03-09
Fix overriding default environment variables on macOS

benj0c:
This pull requests fix the issue with overriding default `UID` and `GID` environment variables on  macOS #8


shiroyasha:
Nice! Thank you for fixing this problem.

PR #11, opened 2023-03-14
Introduce People, Accounts, and Uberauth

shiroyasha:
This pull request introduces the functionality of `People` and `Accounts` in our application. The new functionality is based on the newly introduced `phx.gen.auth` generator from Phoenix and good old Uberauth that allows SSO with multiple providers such are Google, Azure, GitHub and a bunch of others.

The current implementation offers a single way to sign in, via the "Sign in with Google" button.

## Why we have separate Person and Accounts models?

The decision to create separate models for `People.Person` and `People.Accounts` was made to offer modeling people in an organization without strictly requiring everyone to log in first. In other words, users of the system are able to create people records and bootstrap their organisational schema before inviting the rest of the organization.

While the Person model is responsible for storing personal information such as name, email address, and title in the organization, the Accounts model is responsible for storing login credentials, oauth tokens, and scopes.

In addition to providing a way to model the organisation in advance, the separation also offers better security isolation between these two concerns.

## Testing 

In unit tests, the log in flow is provided by the `register_and_log_in_account` and `log_in_account`. To us these two hooks add the following to the top of your tests:

``` elixir
use OperatelyWeb.ConnCase

setup :register_and_log_in_account
```

In Wallaby, browser based tests, the login functionality is provided via the special controller action under the `AccountOauthController` named `test_login`.

The `test_login` functionality is compiled only in the test environment, and provides a convenient hook to log in to the application by visiting `/accounts/auth/test_login` and providing the auth information via the query parameters, ex. `email=john@johnson.com` which we use in out tests.

To use this step in Gherkin specification, start your scenario with the following stanza:

``` cucumber
Given I am logged in as a user
```

This functionality, the routes and the controller action only work if the `config :operately, :test_routes` is set to `true`. We are adding this config in the `config/test.exs` file.

## Google Login and Google Application

To facilitate local development, I've created a Google Application under `console.google.com`.
The process to do it is the following:

- Go to API and Services
- Go to OAuth consent screen
- Create a new application

Then, go to the Credentials section:

- Create a new OAuth 2.0 Client IDs
- Under the callback enter `http://localhost:400/accounts/auth/google/callback`
- Add a list of people who are able to access it
- Generate the API keys
- Create a new .env file in the root of the project
- Fill it with the following content:

``` bash
GOOGLE_LOGIN_CLIENT_ID="<id>"
GOOGLE_LOGIN_CLIENT_SECRET="<secret>"
```

The `.env` file will not be committed to the repository as it is added to the `.gitignore` file.

[...]

PR #13, opened 2023-03-16
Application navigation, Basic design elements and Sidebar

shiroyasha:
This pull request introduces navigation elements to the top and sidebar of the web application, along with basic placeholder design elements. While this design is not of high quality, it will suffice for the alpha release.

The design of the navigation elements is simple and clean, with a focus on usability and ease of use. The top navigation bar uses a simple color scheme, with a logo on the left search in the middle. The sidebar navigation uses a collapsible menu, with section headers and sub-links.

In addition to the navigation elements, this pull request includes basic placeholder design elements throughout the application. These design elements are not intended to be the final design, but rather placeholders to be replaced with higher quality designs in future releases.

### Separation of the objectives and key results

This pull request also separates the creation of objectives and key results in order to reduce the mental complexity involved in setting up an objective.

Traditionally, objectives and key results are created together as a single entity, with each key result being tied to a specific objective. However, this approach can be overwhelming and confusing for users, especially when setting up multiple objectives at once.

With this pull request, the creation of objectives and key results are separated into two distinct steps. First, users create their objectives, providing a clear and concise description of the desired outcome. Then, they create their key results, which are specific, measurable results that indicate progress towards the objective.

### Preview video

Video on Youtube => 👁️ => 

Pull Request #11: Preview


lib/operately_web/controllers/objective_html/new.html.heex
```diff
@@ -1,62 +1,31 @@
...
-
-  <hr>
-
-  <p>How will you measure success?</p>
-
-  <div id="key-results-container">
+<.form :let={f} for={@changeset} action={~p"/objectives"} class="px-4 py-4">
+  <div class="mb-8 flex justify-between items-center">
+    <div class="text-xl font-bold">
+      Add an Objectives
```
bmarkons, line 4: Add an Objective

shiroyasha: Good catch! <3

PR #17, opened 2023-03-21
Fix macOS development environment setup

benj0c:
In this pull request I'm fixing problem with setup development environment on macOS.
- Fix `groupadd` problem when group already exists
- Use project root directory as default for screenshots

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ USER_CONTEXT = export GROUP_ID=$$(id -g) && export USER_ID=$$(id -u)
 # Most importantly, it is able to seemlesly use ARM images on ARM machines, and
 # AMD64 images on AMD64 machines.
 #
-DOCKER_COMPOSE = docker compose run --rm -v /vagrant/screenshots:/tmp/screenshots
+DOCKER_COMPOSE = docker compose run --rm -v $(PWD)/screenshots:/tmp/screenshots
 
 #
 # Prepare commands to run the containers in various modes.
--- a/Dockerfile
+++ b/Dockerfile
@@ -55,7 +55,7 @@ RUN curl -fsSL https://deb.nodesource.com/setup_18.x | bash - \
 # We will call this user 'dev' inside of the development container.
 # The name might differ from your host machine, but the UID and GID will be the same.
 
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -40,7 +40,7 @@ USER_CONTEXT = export GROUP_ID=$$(id -g) && export USER_ID=$$(id -u)
 # Most importantly, it is able to seemlesly use ARM images on ARM machines, and
 # AMD64 images on AMD64 machines.
 #
-DOCKER_COMPOSE = docker compose run --rm -v /vagrant/screenshots:/tmp/screenshots
+DOCKER_COMPOSE = docker compose run --rm -v $(PWD)/screenshots:/tmp/screenshots
```
shiroyasha, line 43: @benj0c I tried to do this last week, but it didn't work on my machine. Were you able to take screenshots with this?

One way to test it is to add:

Screenshot 2023-03-22 at 09 43 11

and then run tests with `make test FILE=test/features/kpis_test.ext`. A new screenshot would appear in the file if it works.

shiroyasha:
Thank you for the pull-request. <3
I'll try to test it on on my machine as well.

Makefile
shiroyasha: I tried it out locally, unfortunately I get this message:

Screenshot 2023-03-22 at 17 22 55

and there are no screenshots in the folder.

shiroyasha: The error message is comming from here: https://github.com/elixir-wallaby/wallaby/blob/main/lib/wallaby/browser.ex#L247-L260.

Unfortunately, the code ignores the error message and simply prints this error.
I'll try to fork and modify the code.

shiroyasha: Some access error for the folder. I assume the `dev` in the container has no write access to it.

Screenshot 2023-03-22 at 17 28 33

shiroyasha: I figured it out:

If I just download the source code, and run `make test FILE=...` it will cause this problem, because docker will create a new folder owned by root on my machine.

However, If I do:

```
mkdir screenshots
make test FILE=...
```

I explicitly create the correct folder on the machine with my development user, and then it gets properly mounted in the docker image.

PR #473, opened 2024-05-21
Removing Notification button on Personal Settings

Jh0wjso:
**Objective**

- Remove the Notifications button in the Personal Settings area.

**Changes Implemented**

- Removed the Notifications button from the Personal Settings interface.
- Adjusted the layout to ensure consistency after the button removal.

### DIFF
--- a/assets/js/pages/AccountPage/page.tsx
+++ b/assets/js/pages/AccountPage/page.tsx
@@ -22,7 +22,7 @@ export function Page() {
           <div className="bg-surface-dimmed rounded-lg overflow-hidden divide-y divide-surface-outline border border-surface-outline">
             <ProfileLink />
             <AppearanceLink />
-            <NotificationLink />
+            {/*<NotificationLink />*/}
           </div>
 
           <div className="bg-surface-dimmed rounded-lg overflow-hidden border border-surface-outline">
--- a/assets/js/pages/AccountPage/page.tsx
+++ b/assets/js/pages/AccountPage/page.tsx
@@ -22,7 +22,6 @@ export function Page() {
           <div className="bg-surface-dimmed rounded-lg overflow-hidden divide-y divide-surface-outline border border-surface-outline">
             <ProfileLink />
             <AppearanceLink />
-            {/*<NotificationLink />*/}
           </div>
 
           <div className="bg-surface-dimmed rounded-lg overflow-hidden border border-surface-outline">
--- a/test/features/notifications_test.exs
+++ b/test/features/notifications_test.exs
@@ -13,7 +13,7 @@ defmodule Operately.Features.NotificationsTest do
     ctx = Map.put(ctx, :reviewer, person_fixture_with_account(%{company_id: ctx.company.id, full_name: "John Reviewer"}))
     ctx = Map.put(ctx, :group, group_fixture(ctx.champion, %{company_id: ctx.company.id, name: "Designers"}))
 
-    ctx = UI.login_as(ctx, ctx.reviewer) # Corrigido para login como revisor
+    ctx = UI.login_as(ctx, ctx.reviewer)
 
     {:ok, ctx}
   end
@@ -21,7 +21,7 @@ defmodule Operately.Features.NotificationsTest do
   feature "unread notifications count", ctx do
     ctx
     |> given_a_project_creation_notification_exists()
-    |> NotificationsSteps.assert_notification_count(1) # Removido login_as(ctx.champion)
+    |> NotificationsSteps.assert_notification_count(1)
     |> NotificationsSteps.visit_notifications_page()
     |> NotificationsSteps.click_on_first_notification()
     |> NotificationsSteps.assert_no_unread_notifications()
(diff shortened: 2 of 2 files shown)


assets/js/pages/AccountPage/page.tsx
```diff
@@ -22,7 +22,7 @@ export function Page() {
           <div className="bg-surface-dimmed rounded-lg overflow-hidden divide-y divide-surface-outline border border-surface-outline">
             <ProfileLink />
             <AppearanceLink />
-            <NotificationLink />
+            {/*<NotificationLink />*/}
```
shiroyasha, line 25: Delete it. When we need it, we will find it in Git history.

test/features/notifications_test.exs
```diff
@@ -6,21 +6,22 @@ defmodule Operately.Features.NotificationsTest do
...
-  
+
   setup ctx do
     ctx = Map.put(ctx, :company, company_fixture(%{name: "Test Org"}))
     ctx = Map.put(ctx, :champion, person_fixture_with_account(%{company_id: ctx.company.id, full_name: "Dorcy Devonshire"}))
     ctx = Map.put(ctx, :reviewer, person_fixture_with_account(%{company_id: ctx.company.id, full_name: "John Reviewer"}))
     ctx = Map.put(ctx, :group, group_fixture(ctx.champion, %{company_id: ctx.company.id, name: "Designers"}))
 
-    UI.login_as(ctx, ctx.reviewer)
+    ctx = UI.login_as(ctx, ctx.reviewer) # Corrigido para login como revisor
```
shiroyasha, line 16: No need for comments on this line, the action is self-evident.

test/features/notifications_test.exs
```diff
@@ -6,21 +6,22 @@ defmodule Operately.Features.NotificationsTest do
...
+
+    {:ok, ctx}
   end
 
   feature "unread notifications count", ctx do
     ctx
     |> given_a_project_creation_notification_exists()
-    |> UI.login_as(ctx.champion)
-    |> NotificationsSteps.assert_notification_count(1)
+    |> NotificationsSteps.assert_notification_count(1) # Removido login_as(ctx.champion)
```
shiroyasha, line 24: Same here, no need for a comment.

test/features/notifications_test.exs
```diff
@@ -6,20 +6,21 @@ defmodule Operately.Features.NotificationsTest do
...
-    UI.login_as(ctx, ctx.reviewer)
+    ctx = UI.login_as(ctx, ctx.reviewer)
+
+    {:ok, ctx}
   end
 
   feature "unread notifications count", ctx do
     ctx
     |> given_a_project_creation_notification_exists()
-    |> UI.login_as(ctx.champion)
```
shiroyasha, line 22: I think the test won't work if you remove this line. What result do you get locally when you run it?

Jh0wjso: After running the tests. During compilation, I noticed warnings indicating that some variables were declared but not used in different parts of the code. For example, in `lib/operately/activities/notifications/task_assignee_assignment.ex`, a variable named `activity` was unused in the function `Operately.Activities.Notifications.TaskAssigneeAssignment.dispatch/1`, and the same occurred in other similar files.

Additionally, there were warnings about imports that were not being used anywhere in the code. For instance, in `lib/operately_email/emails/task_adding_email.ex`, there was an import of `OperatelyEmail.Mailers.ActivityMailer` that was not utilized.

However, the process was interrupted due to a specific error during the test execution in a file named `project_creation_steps.ex`. The error indicated that it was not possible to use the `ExUnit.Case` module without properly starting the ExUnit application. This suggests that there may be a configuration or definition issue in the test support file `project_creation_steps.ex`.

shiroyasha: > However, the process was interrupted due to a specific error during the test execution in a file named project_creation_steps.ex. The error indicated that it was not possible to use the ExUnit.Case module without properly starting the ExUnit application. This suggests that there may be a configuration or definition issue in the test support file project_creation_steps.ex.

Do you get this same problem if you run the test with the following command?

```
make test FILE=test/features/notifications_test.exs:21
```

If yes, please paste the output here so we can analyze it.

Jh0wjso: No, I didn't have the same problem, I just get a test that failed, because I removed the button, I've made a summary below:

Failed Test: The test for unread notifications count (test/features/notifications_test.exs:21) failed.

**Error Details:**

- Expectation: The test expected to find one visible element with the CSS selector [data-test-id="unread-notifications-count"].
- Actual Result: No visible elements matched this selector.

**Stack Trace:**

- The error originated from Operately.Support.Features.UI.assert_has/2 in test/support/features/ui.ex:122.
- This led to the failure in Operately.Support.Features.NotificationsSteps.assert_notification_count/2 in test/support/features/notifications_steps.ex:28.
- The specific test call was in test/features/notifications_test.exs:24.
- 

**Summary of Test Results:**

- Total time taken: 10.4 seconds (all synchronous).
- Features tested: 2
- Failures: 1
- Excluded: 1

PR #475, opened 2024-05-22
Fix flaky projects test

Rockyy174:
Improve the following flaky tests:

"feature connect a goal to a project"
"feature adding a project contributor"

The tests were performing actions in the UI and testing if the actions had written data to the database correctly. They were failing because they database checks were happening before the previous actions had been completed.


shiroyasha:
Looks good!

PR #476, opened 2024-05-23
docs: Self-Hosted Installation docs

shiroyasha:
This is a proposal how the installation steps should look like. For the first pass, we will release Operately that is able to run on a single host bundled with a database and file storage.

The missing part for this installation is setting up a valid SSL cert. Setting this up is probably the most complicated step, as it either requires setting up a letsencrypt schema, or purchasing an SSL cert. Not yet sure how to tackle that one. 

**What is not included in this PR?**

HA configuration will come in the future. The installation steps for HA are more involved, as you need to set up an external database, external S3 based file storage, and a load balancer. The HA steps will most likely be based on Kubernetes manifests. 

---

@markoa @Rockyy174 I'm adding you as reviewers to this PR. My main question is your overall opinion about the approach taken in the docs. Are they understandable? Do you feel that we should automate some parts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,11 @@ organization with 10x less effort.
 
 This project is in alpha and may experience significant changes.
 
+## Installation
+
+- [Try out Operately on your local machine](docs/installation/local.md)
+- [Single Host Installation](docs/installation/single-host.md)
+
 ## Development
 
 - [How to contribute to Operately](CONTRIBUTING.md)
--- a/README.md
+++ b/README.md
@@ -43,8 +43,9 @@ This project is in alpha and may experience significant changes.
 
 ## Installation
 
-- [Try out Operately on your local machine](docs/installation/local.md)
-- [Single Host Installation](docs/installation/single-host.md)
+- [Try out Operately Locally](docs/installation/local.md): Explore Operately on your local machine.
+- [Single Host Installation](docs/installation/single-host.md): Perfect for teams up to 30 people.
+- [Kubernetes Installation](docs/installation/kubernetes.md): For larger teams and enterprise use cases (coming soon).
 
 ## Development
 
--- a/docs/installation/single-host.md
+++ b/docs/installation/single-host.md
@@ -6,7 +6,7 @@ Pre-requisites:
 
 - [Pick a server](#pick-a-server)
 - [Point a domain to your server](#point-a-domain-to-your-server)
-- [Set up a mailbox](#set-up-a-mailbox)
+- [Set up mail server](#set-up-mail-server)
 
 Installation steps:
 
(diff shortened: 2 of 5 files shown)


markoa:
Re. SSL, have you looked into certbot? [ChatGPT recommends](https://chatgpt.com/share/28696a2e-50e1-4d91-9a03-60038a748ff3).

docs/installation/single-host.md
```diff
@@ -0,0 +1,92 @@
+# Instal Operately on a single host
+
+This guide will walk you through the steps to install Operately on a single host.
+
+Pre-requisites:
+
+- [Pick a server](#pick-a-server)
+- [Point a domain to your server](#point-a-domain-to-your-server)
+- [Set up a mailbox](#set-up-a-mailbox)
```
markoa, line 9: A detail but "mailbox" feels kind of misleading, I interpret it more like a thing to _receive_ emails rather than send. Maybe just say "mail server".

docs/installation/single-host.md
```diff
@@ -0,0 +1,92 @@
...
+Operately uses environment variables to configure the application. Edit the `operately.env` and
+fill in the required values:
+
+```bash
+OPERATELY_HOST="operately.example.com"
+SENDGRID_API_KEY="your-sendgrid-api-key"
+OPERATELY_BLOB_TOKEN_SECRET="generate-a-random-string"
+```
+
+To generate a random string for `OPERATELY_BLOB_TOKEN_SECRET`, you can use the following command:
```
markoa, line 77: This is something that we can automate in the future.

It seems like everything from L52 onward can be wrapped in a magic installation script that asks you for the host and api key(s).

markoa (approved):
I think this is a great start. There's room for futher reduction via automation but it's not something to sweat about too much while in alpha imo.

README.md
```diff
@@ -45,6 +45,11 @@ organization with 10x less effort.
 
 This project is in alpha and may experience significant changes.
 
+## Installation
+
+- [Try out Operately on your local machine](docs/installation/local.md)
+- [Single Host Installation](docs/installation/single-host.md)
```
markoa, line 51: btw I understand this will not result in a nuclear attack proof installation :D but it is what it is. I think we should use a more user friendly language like "How to self-host Operately".

shiroyasha:
Thanks for the review. 

Certbot approach does look good 👍🏻, but it requires the installation of Nginx on the host, which is not the optimal Phoenix installation. Hoewever, the line of thinking might be applicable to Phoenix as well.

README.md
shiroyasha: I wanted to leave room for other types of installations. Maybe you can help me with the language, this is what I want:

```
## Installation

- Try out Operately on your local machine
- Single Host installation (good for teams up to 30 members)
- Kubernetes based installation (for HA and bigger teams)
```

This is what GPT suggests:

Screenshot 2024-05-24 at 12 41 05

shiroyasha:
Re: Certbot x Phoenix apps. Found this built-in SSL cert manager written by Sasa Juric which looks promising. https://github.com/sasa1977/site_encrypt

markoa:
respect for calling all web apps "sites" 😂

shiroyasha:
@Rockyy174 I've added the docker-compose and operately.env templates that we can use for testing out the installation and first-time setups.

Note for the future: The `operately.env` will benefit from a couple of iterations of simplifications. The current one contains repeated values and a non-uniform naming scheme, e.g. `POSTGRES_PASSWORD` and `DB_USERNAME`.

shiroyasha:
Small changes implemented in the PR, larger ones will wait for upcoming pull-requests and will be based on installation drilling.

PR #480, opened 2024-05-23
Visual changes for timezones

shiroyasha:
Replaces "Actual Timezone" with "Timezone".
Removes underscore from some timezone labels, e.g. "Sao_Paulo" is now "Sao Paulo".


Jh0wjso:
Looks perfect. Merging.

PR #482, opened 2024-05-24
First time setup

Rockyy174:
I've created a wizard for the self-hosted installations. When the user runs the server for the first time, they are redirected to the /first-time-setup page, where they can create their company and their admin account. After going through the first time setup, they are redirected to the homepage, and they can start using Operately normally.

To make sure that the user goes through the first time setup only if they haven't set up their company and admin account yet, I've created a new plug, `redirect_if_account_not_setup`, which takes care of two tasks:

  - It always redirects users to /first-time-setup if they haven't set up their account yet.
  - It always redirects users from /first-time-setup to the homepage if they have already set up their account.

However, there remains an issue to be solved. In order to allow the necessary unauthenticated requests through, I had to make some changes in the `require_authenticated_account` plug. Unfortunately, these changes broke a couple of tests in test/operately_web/account_auth_test.exs and I haven't figured out a way to adapt the tests yet. 

This test suite uses a custom `conn` which seems not to go through `Plug.Parsers`, therefore 3 of the test cases throw the following error:

```
** (ArgumentError) cannot fetch key "operationName" from conn.params because they were not fetched. 
Configure and invoke Plug.Parsers to set params based on the request
```

### DIFF
--- a/lib/operately/operations/company_adding.ex
+++ b/lib/operately/operations/company_adding.ex
@@ -1,0 +1,32 @@
+defmodule Operately.Operations.CompanyAdding do
+  alias Ecto.Multi
+  alias Operately.Repo
+  alias Operately.Companies.Company
+  alias Operately.People.Account
+  alias Operately.People.Person
+
+  def run(attrs) do
+    Multi.new()
+    |> Multi.insert(:company, Company.changeset(%{name: attrs.company_name}))
+    |> Multi.insert(:account, Account.registration_changeset(%{email: attrs.email, password: attrs.password}))
+    |> insert_person(attrs.full_name, attrs.role)
+    |> Repo.transaction()
+    |> Repo.extract_result(:company)
+  end
+
+  def insert_person(multi, full_name, role) do
+    Multi.run(multi, :person, fn _repo, %{company: company, account: account} ->
+      Person.changeset(%{
+        company_id: company.id,
+        account_id: account.id,
+        full_name: full_name,
+        email: account.email,
+        avatar_url: "",
+        title: role
+      })
+      |> Repo.insert()
+
+      {:ok, nil}
+    end)
+  end
+end
--- a/lib/operately/operations/company_adding.ex
+++ b/lib/operately/operations/company_adding.ex
@@ -4,29 +4,43 @@ defmodule Operately.Operations.CompanyAdding do
   alias Operately.Companies.Company
   alias Operately.People.Account
   alias Operately.People.Person
+  alias Operately.Groups.Group
 
   def run(attrs) do
     Multi.new()
     |> Multi.insert(:company, Company.changeset(%{name: attrs.company_name}))
+    |> insert_group()
     |> Multi.insert(:account, Account.registration_changeset(%{email: attrs.email, password: attrs.password}))
     |> insert_person(attrs.full_name, attrs.role)
     |> Repo.transaction()
     |> Repo.extract_result(:company)
   end
 
-  def insert_person(multi, full_name, role) do
-    Multi.run(multi, :person, fn _repo, %{company: company, account: account} ->
-      Person.changeset(%{
-        company_id: company.id,
-        account_id: account.id,
-        full_name: full_name,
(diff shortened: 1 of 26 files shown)


lib/operately/operations/company_adding.ex
```diff
@@ -0,0 +1,46 @@
+defmodule Operately.Operations.CompanyAdding do
+  alias Ecto.Multi
+  alias Operately.Repo
+  alias Operately.Companies.Company
+  alias Operately.People.Account
+  alias Operately.People.Person
+  alias Operately.Groups.Group
+
+  def run(attrs) do
```
shiroyasha, line 9: We need to check here if it is still ok to run this operation to avoid the possibility of malicious already singed in users manually crafting a graphql query and creating new records in the database.

Rockyy174: I've made some changes and I believe it's okay now. I also added some tests to make sure the endpoint doesn't accept requests if the account has already been setup.

Rockyy174:
I've made some changes following the suggestions in https://github.com/operately/operately/pull/483#issue-2315665947
After the previous changes, I think `OperatelyWeb.AccountAuth` had gotten a bit messy. In my opinion, it's better now since public/private operations and pages are handled by `OperatelyWeb.PageController` and `OperatelyWeb.Graphql.Context`.
The code seems to be working well and all tests are passing.

There remains one issue to be solved. Although the tests are passing, the changes in the code cause some requests in the tests to get 401 responses and, as a result, many tests are showing error messages in the terminal when they are run locally. 

It seems that the error happens when this function is called:

https://github.com/operately/operately/blob/main/test/support/features/notifications_steps.ex#L17-L23
````
  def assert_notification_exists(ctx, author: author, subject: subject) do
    ctx
    |> UI.sleep(100)
    |> visit_notifications_page()
    |> UI.assert_text(author.full_name)
    |> UI.assert_text(subject)
  end

````

I also noticed that whenever `assert_notification_exists` fails, it's because both GraphQL requests to "GetMe" and "UnreadCount" fail. For some reason, these two requests don't have `:current_account` assigned to them, but I couldn't figure out why.

File: lib/operately_web/graphql/context.ex [hunk 2]
```diff
@@ -1,10 +1,48 @@
...
-    Absinthe.Plug.put_options(conn, context: context)
+      Absinthe.Plug.put_options(conn, context: context)
+    else
+      #
+      # IOs added for debugging the current errors in the tests
+      # They will be removed after the errors are fixed
+      #
+      IO.puts("\nFail")
+      IO.inspect(conn.params)
+      IO.puts("\n")
```
Rockyy174, line 44: I ended up adding these `IO`s here for debugging purposes, but after we solve the 401 problems with the tests, this code should definitely be removed.

shiroyasha:
Just to confirm, this is the error that you are seeing
https://operately.semaphoreci.com/jobs/912d6ef9-97b6-414f-b0d9-e2d7534ad373#L1965 ?

Rockyy174:
> https://operately.semaphoreci.com/jobs/912d6ef9-97b6-414f-b0d9-e2d7534ad373#L1965

Yes, that's the error.

shiroyasha:
I've investigated the problem a bit more.

It seems that the problem was already present on the main branch as well, just the format was different as we were not returning a 401 JSON error. https://operately.semaphoreci.com/jobs/f75b3c40-be0e-48b5-8f57-6a521866c7f6#L2377

There is a good chance that it no longer exists after this PR has been merged https://github.com/operately/operately/pull/481. For example, I don't see it anymore on the latest build on the main branch.

---

Given that this problem was already present on the main branch, it means that it was not caused by the code in this pull-request. I propose to merge this PR (including the IO logs), and if the problem persist we will resolve it a dedicated pull-request.

What do you think @Rockyy174, are you ok with me merging this PR?

Rockyy174:
I've just added a new test for the first-time-setup form and fixed a bug in the redirecting logic. I think we can merge it more confidently now.

PR #490, opened 2024-05-28
Add site_encrypt configuration

Rockyy174:
I've added the configuration for [site_encrypt](https://github.com/sasa1977/site_encrypt) as suggested in https://github.com/operately/operately/pull/476#issuecomment-2129436176.

Unfortunately, there were some conflicts with our current configuration and I didn't manage to make it work. In the error message ([error-message.txt](https://github.com/operately/operately/files/15470123/error-message.txt)), the following lines stood out to me:

```
[notice] Application operately exited: Operately.Application.start(:normal, []) returned an error: shutdown: failed to start child: SiteEncrypt.Phoenix.Endpoint
** (EXIT) an exception was raised:
** (RuntimeError) could not find persistent term for endpoint OperatelyWeb.Endpoint. Make sure your endpoint is started and note you cannot access endpoint functions at compile-time
```

It seems the problem is in our endpoint. In the site_encrypt's docs, they tell us to add the following lines to the endpoint and application:

```
defmodule PhoenixDemo.Endpoint do
   use SiteEncrypt.Phoenix.Endpoint, otp_app: :my_app
   # ...
end

defmodule PhoenixDemo.Application do
  def start(_type, _args) do
    children = [PhoenixDemo.Endpoint]
 
  # ...
end
```

But because we are using Absinthe, our configuration is slightly different:

```
defmodule OperatelyWeb.Endpoint do
  use SiteEncrypt.Phoenix.Endpoint, otp_app: :operately
  use Absinthe.Phoenix.Endpoint
 # ...
end

defmodule Operately.Application do
   # ...

  def start(_type, _args) do
    children = [
      OperatelyWeb.Endpoint,
      {Absinthe.Subscription, OperatelyWeb.Endpoint},

   # ...
end
```

I think the problem may be related to this, since this is the main difference between our configuration and the documentation.
It's worth mentioning that I created a dummy Phoenix app and configured it exactly as the docs showed and it worked. So, that confirms that the problem is specific to our configuration.


shiroyasha:
Closing in favor of https://github.com/operately/operately/pull/500.

PR #505, opened 2024-06-03
Allow admin accounts to invite members

Rockyy174:
Now an admin account can invite new member accounts to join the company. They can do it by issuing a URL with a token which the new member can use to log in for the first time and reset their password.

We still have to work on some improvements, though:

- Copying the URL manually is not a good experience. We could add a "copy to clipboard" button to make the admin's life easier.
- The tokens are valid for only 1 hour, so probably there should be an option for the admin to reissue tokens.
- Regarding the notifications, I only added a placeholder to suppress the errors. Maybe proper notifications could be configured.


lib/operately/invitations/invitation_token.ex
```diff
@@ -0,0 +1,61 @@
+defmodule Operately.Invitations.InvitationToken do
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  @primary_key {:id, :binary_id, autogenerate: true}
+  @foreign_key_type :binary_id
```
shiroyasha, line 6: You can use `use Operately.Schema` defined in lib/operately/schema.ex to reduce the boilerplate:

Before:

```
  use Ecto.Schema
  import Ecto.Changeset

  @primary_key {:id, :binary_id, autogenerate: true}
  @foreign_key_type :binary_id
```

After:

```
  use Operately.Schema
```

PR #518, opened 2024-06-06
Admins can reissue invitation tokens

Rockyy174:
Admin users can now reissue invitation tokens to new members. When a new token is issued, the old one is automatically deleted. New tokens are valid for 1 hour.

Before merging this PR, we may want to consider the following:

Currently, the admin will see the option to reissue invitation tokens for **every** member **forever**. I think that may be annoying to admins because:

- Once a member uses the token and sets up their password, it doesn't make much sense for the admin to reissue a new token. Even if the admin were to issue a token for the member so they can reset their password, I guess that shouldn't be an **invitation** token.

- Tokens cannot be reissued for members who did not join the company via an invitation token since there is no invitation associated with such members. 
(I made sure to handle this exception, so if the admin tries to reissue a token for such members, they will see a message indicating that it's not possible.)

Having said that, the way to fix these issues would be to display the "reissue token" option only for members who have joined the company via an invitation token and have not reset their passwords yet. 
The most straightforward solution would be to add a flag/field to the person schema to indicate whether a member might need a new token in the future. But maybe we can come up with a better way to handle this.

Now, this problem only exists if we assume that having the "reissue token" option forever available for every member would bother admins. If we conclude that admin users wouldn't mind it, then the PR should be ready.

### DIFF
--- a/assets/js/pages/CompanyAdminManagePeoplePage/RemoveMember.tsx
+++ b/assets/js/pages/CompanyAdminManagePeoplePage/RemoveMember.tsx
@@ -1,0 +1,60 @@
+import React, { useState } from "react";
+import { useNavigate } from "react-router-dom";
+
+import { useRemoveMemberMutation } from "@/models/companies";
+import { createTestId } from "@/utils/testid";
+import { Person } from "@/gql";
+import { FilledButton, GhostButton } from "@/components/Button";
+import Modal from "@/components/Modal";
+
+
+
+export default function RemoveMember({person}: {person: Person}) {
+  const removeTestId = createTestId("remove", person.fullName);
+  const [showRemoveMember, setShowRemoveMember] = useState(false);
+  
+  const navigate = useNavigate();
+  const [remove, { loading }] = useRemoveMemberMutation({
+    onCompleted: () => {
+      navigate(0);
+    }
+  });
+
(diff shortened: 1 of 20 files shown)


shiroyasha:
We can merge it as it is, but please write down the proposed improvements as an issue, both proposal make complete sense. 👍

assets/js/pages/CompanyAdminManagePeoplePage/RemoveMember.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+
+export default function RemoveMember({person}: {person: Person}) {
+  const removeTestId = createTestId("remove", person.fullName);
+  const [showRemoveMember, setShowRemoveMember] = useState(false);
+  
+  const navigate = useNavigate();
+  const [remove, { loading }] = useRemoveMemberMutation({
+    onCompleted: () => {
+      navigate(0);
```
shiroyasha, line 19: What is `navigate(0)`?

Rockyy174: When a member is removed, the list of people in the page has to be refreshed. `navigate(0)` was just a way I found to reload the current route.
I probably should've added a comment since this is definitely not intuitive :sweat:

PR #524, opened 2024-06-07
[Draft] Add schemas and migrations for new security model

Rockyy174:
I've added the new schemas and migrations for:

- Access Bindings
- Access Contexts
- Access Groups
- Access Group Memberships

What do you think, @shiroyasha?


lib/operately/access/access_binding.ex
```diff
@@ -0,0 +1,24 @@
+defmodule Operately.Access.AccessBinding do
+  use Ecto.Schema
```
shiroyasha, line 2: Use Operately.Schema

lib/operately/access/access_binding.ex
```diff
@@ -0,0 +1,24 @@
...
+  import Ecto.Changeset
+
+  @primary_key {:id, :binary_id, autogenerate: true}
+  @foreign_key_type :binary_id
+  schema "access_bindings" do
+    belongs_to :access_group, Operately.Access.AccessGroup
+    belongs_to :access_context, Operately.Access.AccessContext
+    field :access_level, Ecto.Enum, values: [:full_access, :edit_access, :comment_access, :view_access]
```
shiroyasha, line 10: Let's use a integer for this field, that way we can run queries like:

```
access_level > ^Operately.Access.Level.view_access()
```

For the values, please use:

```
view - 10
comment - 40
edit - 70
full - 100
```

I'm spreading them out to leave space for additional levels if we need them in the future.

lib/operately/access/access_binding.ex
```diff
@@ -0,0 +1,24 @@
+defmodule Operately.Access.AccessBinding do
```
shiroyasha, line 1: Can we call it `Operately.Access.Binding` to avoid the double Access in the name?

lib/operately/access/access_binding.ex
```diff
@@ -0,0 +1,24 @@
...
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  @primary_key {:id, :binary_id, autogenerate: true}
+  @foreign_key_type :binary_id
+  schema "access_bindings" do
+    belongs_to :access_group, Operately.Access.AccessGroup
```
shiroyasha, line 8: Let's use `belongs_to :group` and `belongs_to :context` if it is possible.

lib/operately/access/access_binding.ex
Rockyy174: Sure. 
How about AccessContext, AccessGroup and AccessGroupMembership? Should I remove the prefix "Access" as well?

shiroyasha: Yes.

Rockyy174:
I removed the prefixes `Access` and `access_` everywhere, except for the table names. Once I add the unit tests, we should find out whether I forgot to remove the prefixes in some place.
Let me know if there is any improvement we can make, @shiroyasha.
If everything is fine, then I will go ahead and start adding the one-to-one relationships with projects, spaces, activities and organizations.

PR #529, opened 2024-06-07
Generate hooks for queries

shiroyasha:
Continuation of https://github.com/operately/operately/pull/528.

Draft because I still need to figure out what is the best implementation for the hooks. Should we use a library, or just create a hook ourselves. 🤔


Rockyy174:
Do you have any specific library in mind, @shiroyasha?

shiroyasha:
So far, not really. I want to look into what popular options are out there.

PR #541, opened 2024-06-10
Add access context to projects

Rockyy174:
I've added a one-to-one relationship between projects and access contexts.
I've also added a data migration to create access contexts for existing projects.


shiroyasha (approved):
Looks good!

PR #553, opened 2024-06-11
Creating a project automatically creates access context

Rockyy174:
I changed the project creation logic so that creating a project will automatically create an access context from now on.

### DIFF
--- a/test/features/project_creation_test.exs
+++ b/test/features/project_creation_test.exs
@@ -12,6 +12,7 @@ defmodule Operately.Features.ProjectCreationTest do
     |> Steps.start_adding_project()
     |> Steps.submit_project_form(params)
     |> Steps.assert_project_created(params)
+    |> Steps.assert_access_context_created(params)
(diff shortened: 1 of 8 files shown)


test/features/project_creation_test.exs
```diff
@@ -61,6 +65,7 @@ defmodule Operately.Features.ProjectCreationTest do
     |> Steps.start_adding_project()
     |> Steps.submit_project_form(params)
     |> Steps.assert_project_created(params)
+    |> Steps.assert_access_context_created(params)
```
shiroyasha, line 68: This is probably a bit too low level for feature tests. We can test this in the create project operation instead.

Rockyy174: Makes sense. 
Do you think the unit tests I added are enough? Or do you suggest that I add more?

shiroyasha: Unit tests are enough 👍

PR #557, opened 2024-06-11
Button for Avatar Picture Upload

Jh0wjso:
**Pull Request Description:**

This pull request introduces a fix to prevent a possible "Null Object" error that occurred when attempting to upload files on the user profile page.

**Proposed Changes:**

- In the `ProfileForm.js` file, in addition to the fix for preventing the "Null Object" error, a button to remove the user's avatar has been added.
- The "Remove Avatar" button allows users to remove their current profile image by setting the avatar URL to  `null`.

**Tests Performed:**
- [ ] Manually tested on the user profile page.
- [ ] Verified the absence of errors in the browser console.
- [ ] Confirmed the functionality of avatar upload and removal without occurrence of errors.

Please review and test the proposed changes and share any additional feedback or concerns.

### DIFF
--- a/assets/js/pages/AccountEditProfilePage/index.tsx
+++ b/assets/js/pages/AccountEditProfilePage/index.tsx
@@ -11,10 +11,18 @@ import { S3Upload } from "@/components/Editor/Blob/S3Upload/S3Upload";
 import { CreateBlob } from "@/graphql/Blobs";
 import moment from "moment-timezone";
 import { FilledButton } from "@/components/Button";
+import classnames from "classnames";
 
 export async function loader() {
   return null;
 }
+
+const dimmedClassName = classnames(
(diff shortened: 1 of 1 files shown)


assets/js/pages/AccountEditProfilePage/index.tsx
```diff
@@ -50,16 +51,17 @@ function FileInput({ label, onChange, error }) {
         {label}
       </label>
       <div className="flex-1">
+        <FilledButton type="secondary" onClick={() => document.getElementById("formFileLg")?.click()}>
```
shiroyasha, line 54: Can we do this with React refs instead?

Jh0wjso: True, yes we can, I had forgotten, it actually makes more sense.

PR #570, opened 2024-06-12
Preserve storage type

Jh0wjso:
This PR aims to fix an issue in the create_blob/1 test within the Operately.BlobsTest module. The test was failing due to the absence of the storage_type column in the blobs table in the database.

To address this, this PR adds a migration to add the storage_type column to the blobs table. The migration was created using the mix ecto.gen.migration command and adds the storage_type column with the string type to the blobs table.

Additionally, the create_blob/1 test in the blobs_test.exs file was adjusted to include the storage_type field in the valid attributes passed for blob creation. This ensures that the test executes correctly after the migration is applied.

### DIFF
--- a/test/support/fixtures/blobs_fixtures.ex
+++ b/test/support/fixtures/blobs_fixtures.ex
@@ -12,7 +12,8 @@ defmodule Operately.BlobsFixtures do
       attrs
       |> Enum.into(%{
         filename: "some filename",
-        status: :pending
+        status: :pending,
+        storage_type: :local  # Adiciona storage_type padrão
       })
       |> Operately.Blobs.create_blob()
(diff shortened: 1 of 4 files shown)


test/support/fixtures/blobs_fixtures.ex
```diff
@@ -12,7 +12,8 @@ defmodule Operately.BlobsFixtures do
       attrs
       |> Enum.into(%{
         filename: "some filename",
-        status: :pending
+        status: :pending,
+        storage_type: :local  # Adiciona storage_type padrão
```
shiroyasha, line 16: Please use english comments, or remove them. 🧹

PR #580, opened 2024-06-14
Creating a group automatically creates an access context

Rockyy174:
I've made changes to the way groups are created so that now, creating a group automatically creates an access context. To achieve this, I had to solve two problems:

### Different logics for group creation

Groups were being created in three different places:

- lib/operately/groups.ex
- lib/operately/companies.ex
- lib/operately/operations/company_adding.ex

I didn't want to hard code the insertion of access contexts in all these places as it would make the code hard to maintain. Instead, I implemented the group creation logic similarly to the way projects are created. Just as `create_project` delegates the action to `Operately.Operations.ProjectCreation`, `create_group` now delegates the action to `Operately.Operations.GroupCreation`.

Then, I added `insert_group`, which inserts the group and the context in the multi transactions. Now, all group creation operations go through `insert_group`.

### Broken delete_group function

Since creating a group now automatically creates an access context, deleting a group stopped working due to a DB constraint, causing the unit test for the `delete_group` function to fail. I initially updated delete_group to also delete the access context, but I figured that it was a bad idea because, in the near future, the group's access context will also be associated with activities.

However, if I'm not mistaken, the application in the moment does not support deleting a group and this function was not being used anywhere. Therefore, I decided to delete the function and its unit test for now. In the future, if we implement a "deleting group/space" feature, we can address this issue, most likely with a soft delete.

--

Do you think there are better ways to solve these problems, @shiroyasha? Let me know if you have any suggestions for improvement!

### DIFF
--- a/lib/operately/operations/company_adding.ex
+++ b/lib/operately/operations/company_adding.ex
@@ -4,7 +4,7 @@ defmodule Operately.Operations.CompanyAdding do
   alias Operately.Companies.Company
   alias Operately.People.Account
   alias Operately.People.Person
-  alias Operately.Groups.Group
+  alias Operately.Groups
 
   def run(attrs) do
     Multi.new()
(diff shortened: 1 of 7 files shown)


Rockyy174:
The solution for this [test](https://operately.semaphoreci.com/workflows/090da73e-a3c1-4c86-826e-2f35378c957b/summary?pipeline_id=8a01e559-02d2-4609-bd61-0d86ed512c45&report_id=31c127ba-6edc-33b1-9f53-fc4a54e74c82) is https://github.com/operately/operately/pull/579.

lib/operately/operations/company_adding.ex
```diff
@@ -17,31 +17,31 @@ defmodule Operately.Operations.CompanyAdding do
...
-    end
+    attrs = %{
+      name: "Company",
+      mission: "Everyone in the company",
+      icon: "IconBuildingEstate",
+      color: "text-cyan-500"
+    }
+
+    multi
+    |> Groups.insert_group(attrs)
```
shiroyasha, line 28: Will this generate two activities 1: Company Adding + 2: Space creation?
If yes, I would like to make this to create only one: Company Adding, because the user executed only one action.

Ideally, Operately should correlate "one user action" = "one activity".

shiroyasha:
Regarding this:

> Since creating a group now automatically creates an access context, deleting a group stopped working due to a DB constraint, causing the unit test for the delete_group function to fail. I initially updated delete_group to also delete the access context, but I figured that it was a bad idea because, in the near future, the group's access context will also be associated with activities.
> However, if I'm not mistaken, the application in the moment does not support deleting a group and this function was not being used anywhere. Therefore, I decided to delete the function and its unit test for now. In the future, if we implement a "deleting group/space" feature, we can address this issue, most likely with a soft delete.

It is 👍. Operately should avoid deleting data as this is deleting can delete valuable audit and historical data. When possible, we should opt for soft-delete or archivation, neither of which will delete the access context. 

All of this is for the future, for now, just remove the delete_group on the schema/context level.

lib/operately/operations/company_adding.ex
Rockyy174: If I'm not mistaken, it creates 0 activities. The new function, `insert_group`, doesn't create any activity. It only inserts the group and the context in the `multi` transaction. 

We are using the function `insert_group` in 3 places:

- lib/operately/operations/company_adding.ex
- lib/operately/operations/group_creation.ex
- lib/operately/companies.ex

And in all these places, it seems there is no activity being recorded :neutral_face:

shiroyasha: Heh, we'll have to fix that. Can you open an issue for that so we don't have to fix it in this PR?

We can go forward with this PR then. 0 activities is better than 2 activities.

PR #582, opened 2024-06-17
S3 for rich text attachments

Jh0wjso:
This pull request introduces a new feature that enables the storage of rich text attachments in an S3 bucket based on the environment configuration. This aims to enhance the system's flexibility and scalability when handling rich content.

#### Implementation
The `rich_text/1` function has been modified to support storing attachments in S3 depending on the `OPERATELY_STORAGE_TYPE` environment variable. Here’s a summary of the changes implemented:

1. **Storage Type Check**:
   - The function checks the value of the `OPERATELY_STORAGE_TYPE` environment variable.
   - If the value is `"s3"`, rich text attachments are stored in an S3 bucket specified by the `OPERATELY_S3_BUCKET` environment variable.

2. **S3 Storage Options**:
   - When S3 storage is enabled, the options passed to `Prosemirror2Html.convert/2` include `storage_type: :s3` and the bucket name retrieved from `OPERATELY_S3_BUCKET`.

3. **Fallback to Local Storage**:
   - If the `OPERATELY_STORAGE_TYPE` variable is not set to `"s3"`, the system uses the default local storage.
   - The options include the application’s domain obtained from `OperatelyWeb.Endpoint.url()`.

#### Code

```elixir
def rich_text(content) do
  if System.get_env("OPERATELY_STORAGE_TYPE") === "s3" do
    opts = [storage_type: :s3, bucket: System.get_env("OPERATELY_S3_BUCKET")]
    {:safe, Prosemirror2Html.convert(content, opts)}
  else
    opts = [domain: OperatelyWeb.Endpoint.url()]
    {:safe, Prosemirror2Html.convert(content, opts)}
  end
end
```


assets/js/components/Editor/Blob/FileUploader.tsx
```diff
@@ -19,7 +20,12 @@ export class MultipartFileUploader implements FileUploader {
 
       const signedUploadUrl = blob.data.createBlob.signedUploadUrl;
 
-      await this.uploadFile(file, signedUploadUrl, progressCallback);
+      if (blob.data.createBlob.storageType === "local") {
+        await this.uploadFile(file, signedUploadUrl, progressCallback);
+      }
+      else {
+        await S3Upload(file);
```
shiroyasha, line 27: How come that the S3Upload is not taking progressCallback as an argument? Will the progress bar still work for files uploaded to S3?

lib/operately/blobs.ex
```diff
@@ -31,11 +31,23 @@ defmodule Operately.Blobs do
...
-    host = OperatelyWeb.Endpoint.url()
-    path = "#{blob.company_id}-#{blob.id}"
-    token = Operately.Blobs.Tokens.gen_get_token(path)
+    case Application.get_env(:operately, :storage_type) do
+      "s3" ->
+        host = Application.get_env(:operately, :storage_host)
+        bucket = Application.get_env(:operately, :storage_bucket)
+        path = "#{blob.company_id}-#{blob.id}"
+
+        "http://#{host}:9090/#{bucket}/#{path}"
```
shiroyasha, line 40: The port 9090 seems to be hardcoded here. Amazon S3 will not have this port attached. You will need to handle this in some way.

lib/operately/blobs.ex
```diff
@@ -47,7 +59,9 @@ defmodule Operately.Blobs do
         host = Application.get_env(:operately, :storage_host)
         config = ExAws.Config.new(:s3, host: host)
 
-        ExAws.S3.presigned_url(config, :put_object, bucket, path, expires_in: 3600) |> IO.inspect()
+        ExAws.S3.put_object(bucket, blob.filename, path) |> ExAws.request!()
```
shiroyasha, line 62: I'm confused by this line:

1. It seems that the result is not used on the next line?
2. According to docs, ExAws.S3.put_object is uploading an object to S3, not just creating a signed url. Will this line upload a file every time we call it, or am I missunderstanding what is it doing? 🤔

lib/operately_email/templates.ex
```diff
@@ -41,9 +41,13 @@ defmodule OperatelyEmail.Templates do
   end
 
   def rich_text(content) do
-    opts = [domain: OperatelyWeb.Endpoint.url()]
-
-    {:safe, Prosemirror2Html.convert(content, opts)}
+    if System.get_env("OPERATELY_STORAGE_TYPE") === "s3" do
+      opts = [storage_type: :s3, bucket: System.get_env("OPERATELY_S3_BUCKET")]
+      {:safe, Prosemirror2Html.convert(content, opts)}
```
shiroyasha, line 46: I see that you are passing extra arguments here, but I don't see how they are used in the handler. 

https://github.com/operately/operately/blob/091c9bc8171e7c1d75326fe0305b03921545abd0/lib/prosemirror2html.ex#L89-L93

It seems that it is only using `:domain`, are the extra arguments necessary?

Jh0wjso: This change I made was unnecessary, it was a change that was not supposed to be committed, I will revert it.

lib/operately/blobs.ex
Jh0wjso: I reread the Amazon documentation, and it seems that my code is actually wrong, I tested it locally, and we can leave it as ExAws.S3.presigned_url

PR #587, opened 2024-06-18
Creating a goal automatically creates an access context

Rockyy174:
I've changed the goals creation logic so that, from now on, creating a goal automatically creates an access context.


shiroyasha:
Looks good. Merge?

Rockyy174:
Yes, we can merge it.

PR #610, opened 2024-06-19
[Draft] Add insert_sync to goal operations

Rockyy174:
I've added `insert_sync` to the following operations:

- lib/operately/operations/goal_closing.ex
- lib/operately/operations/goal_reopening.ex
- lib/operately/operations/goal_discussion_creation.ex
- lib/operately/operations/goal_timeframe_editing.ex

But some of the tests that were added in https://github.com/operately/operately/pull/609 don't pass. 
The problem seems to be in the fact that `insert_sync` is triggering the notification dispatch before the activity is updated with the `comment_thread_id`, causing the dispatch function not to work.

Here is what the activity looked like before the operation was changed:

before

And here is after, with `comment_thread_id` as `nil`:

after

Now, this is a bit tricky because the tests themselves are passing. But when the tests are run locally, the errors are visible in the console. 

error

I'm not sure if this problem is fixable with the current approach. I will try another idea that I have and submit it as another draft.
Do you think it's possible to make this approach work, @shiroyasha?

### DIFF
--- a/lib/operately/operations/goal_closing.ex
+++ b/lib/operately/operations/goal_closing.ex
@@ -9,36 +9,31 @@ defmodule Operately.Operations.GoalClosing do
     goal = Goals.get_goal!(goal_id)
 
     changeset = Goals.Goal.changeset(goal, %{
-      closed_at: DateTime.utc_now(), 
+      closed_at: DateTime.utc_now(),
       closed_by_id: author.id,
       success: success
     })
 
     Multi.new()
     |> Multi.update(:goal, changeset)
-    |> Multi.insert(:activity_without_thread, fn _changes ->
-      Activities.Activity.changeset(%{
-        author_id: author.id,
-        action: Atom.to_string(:goal_closing),
-        content: Activities.build_content!(:goal_closing, %{
-          company_id: author.company_id,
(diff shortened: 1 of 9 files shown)


shiroyasha:
My best idea what could be the problem is :inline behaviour of Oban in tests: https://hexdocs.pm/oban/testing.html#setup-testing-helpers.

Basically, it runs the code immediately, instead of how it would run in production and wait for the transaction to finish.

shiroyasha:
This sync execution might also be the source of some of the flakyness, for example here: https://operately.semaphoreci.com/jobs/87ef4167-2d8b-4401-8bfa-8af3a0ddce80/summary?report_id=31c127ba-6edc-33b1-9f53-fc4a54e74c82&test_id=63ed55db-51b2-3bfd-bdff-ebf122754cf2.

Rockyy174:
Yes, running the tests in :manual mode instead of :inline seems to be the way to tackle this problem. 

I made the changes to run the tests in :manual model and now the Oban tasks are successfully not being executed at once and the errors are no longer thrown.

Now, I just have to read the docs and find out how to manually run Oban tasks, since now the tasks are just not running at all.

PR #618, opened 2024-06-20
Add insert_sync to goal operations

Rockyy174:
I've added insert_sync to the following operations:

- lib/operately/operations/goal_closing.ex
- lib/operately/operations/goal_reopening.ex
- lib/operately/operations/goal_discussion_creation.ex
- lib/operately/operations/goal_timeframe_editing.ex

### DIFF
--- a/lib/operately/activities.ex
+++ b/lib/operately/activities.ex
@@ -25,24 +25,34 @@ defmodule Operately.Activities do
         action: action,
         author_id: author_id,
         params: callback.(changes),
-      }) 
+      })
 
       Oban.insert(job)
     end)
   end
 
-  def insert_sync(multi, author_id, action, callback) do
+  def insert_sync(multi, author_id, action, callback, opts \\ []) do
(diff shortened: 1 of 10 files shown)


Rockyy174:
What do you think about this implementation of `insert_sync` and `dispatch_notification`, @shiroyasha?

I tried to do it in a way that It gives us some freedom to call `dispatch_notification` inside or outside `insert_sync`. However, my concern is that the code may have become a bit messy :confused:

Rockyy174:
Also, regarding this [test](https://operately.semaphoreci.com/jobs/e26c36a1-bd6f-403d-bca7-cd403d24d00f/summary?report_id=31c127ba-6edc-33b1-9f53-fc4a54e74c82&test_id=a4eb8551-ac0c-3054-8539-a04de5b0a882), I think it's one of those flaky ones. Locally, it sometimes passes and sometimes doesn't.

PR #619, opened 2024-06-20
Avatar auto reload

Jh0wjso:
This pull request fixes the auto-reload issue of the user avatar when a new avatar is uploaded to the site. Below are the changes made:

1. **Removal of Unused Import:**
   - File: `assets/js/pages/AccountEditProfilePage/index.tsx`
   - Line removed: `import { ImageAvatar } from "@/components/Avatar";`

2. **Modification in the FileInput Component:**
   - Specification of inline style for the file input to ensure it remains hidden.
   - Before: `style={{ display: "none" }}`
   - After: `style={{display: "none"}}`

3. **Update to Avatar Rendering in Profile Form:**
   - Replaced the `ImageAvatar` component with an `` tag to display the avatar image.
   - Added attributes for the image: `src={avatarUrl}`, `alt="avatar"`, `className="rounded-full w-24 h-24"`.

With these changes, the user's avatar is automatically updated on the frontend as soon as a new avatar is uploaded, enhancing the user experience by allowing immediate visual feedback of their changes.

### DIFF
--- a/assets/js/pages/AccountEditProfilePage/index.tsx
+++ b/assets/js/pages/AccountEditProfilePage/index.tsx
@@ -17,7 +17,7 @@ import classNames from "classnames";
 import moment from "moment-timezone";
 
 import { FilledButton } from "@/components/Button";
-import { BackupAvatar, ImageAvatar } from "@/components/Avatar";
+import { BackupAvatar } from "@/components/Avatar";
 
 export async function loader() {
   return null;
@@ -73,7 +73,7 @@ function FileInput({ onChange }) {
           Upload Photo
         </FilledButton>
 
-        <input ref={fileInputRef} onChange={onChange} type="file" accept="image/*" style={{ display: "none" }} />
+        <input ref={fileInputRef} onChange={onChange} type="file" accept="image/*" />
       </div>
     </div>
   );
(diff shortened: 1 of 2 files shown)


assets/js/pages/AccountEditProfilePage/index.tsx
```diff
@@ -169,8 +169,12 @@ function ProfileForm({ me }) {
 
   return (
     <Forms.Form onSubmit={handleSubmit} loading={loading} isValid={isValid}>
-      <section className="flex flex-col w-full justify-center items-center text-center -mt-4">
-        {avatarUrl ? <ImageAvatar person={me} size="xxlarge" /> : <BackupAvatar person={me} size="xxlarge" />}
+      <section className="flex flex-col w-full justify-center items-center text-center mt-4">
+        {avatarUrl ? 
+          <img src={avatarUrl} alt="avatar" className="rounded-full w-24 h-24" />
```
shiroyasha, line 174: I liked the previous version more as it was using the same size standard as the backup avatar. So if the size `xxlarge` changes, it will be changed in the whole application.

With the current solution we are leaving room for future errors in the application.

Can you maybe find a solution that would still use the ImageAvatar?

shiroyasha: Try: `<ImageAvatar person={{avatarUrl}} ...`

PR #624, opened 2024-06-21
Remove deprecated Activities.insert function

Rockyy174:
I have deleted `Operately.Activities.insert` completely from the project and replaced it by `Operately.Activities.insert_sync`.

I am marking this pull request as a draft because I am not 100% sure about some important changes in the code:

- `Operately.Projects.create_review_request` was using `Operately.Activities.insert`, but it wasn't being used anywhere in the project. Furthermore, its activity's action was also deprecated. So, I decided to delete `create_review_request` as well.

- The module `Operately.Activities.Recorder` was only being used by `Operately.Activities.insert`, so I also deleted it.

- `Operately.Activities.insert` was being used in 2 modules that are related to updates: `Operately.Updates` and `Operately.Operations.DiscussionEditing`. 
I spent quite some time trying to get tests to work for these 2 modules before updating them, but the tests just wouldn't work. So we are relying on feature tests, which might not be ideal. If it's decided that we can't afford the risk of relying only on the feature tests, then I should spend some more time trying to get the unit tests to work.

- This pull request depends on some code from [#618](https://github.com/operately/operately/pull/618). So, that pull request would have to be merged first.

If we are okay with the above details, then this pull request should be ready.


lib/operately/operations/task_description_change.ex
```diff
@@ -1,18 +1,31 @@
...
-    changeset = Operately.Tasks.Task.changeset(task, %{description: description})
+    changeset = Task.changeset(task, %{description: description})
+
+    space = from(t in Task,
+      join: m in Operately.Projects.Milestone, on: m.id == t.milestone_id,
+      join: p in Operately.Projects.Project, on: p.id == m.project_id,
+      join: g in Operately.Groups.Group, on: g.id == p.group_id,
+      where: t.id == ^task_id,
+      select: g
+    )
```
shiroyasha, line 19: I think that this one can be implemented in a nicer way. Try to add the following to the task schema:

```
belongs_to :group, through: [:milestone, :project, :group]
```

And then come back here and do:

```
task = Operately.Tasks.get_task!(task_id)
task = Operately.Repo.preload(:group)
```

Rockyy174: I didn't know that was possible. It looks much better now. Thank you!

PR #647, opened 2024-06-27
Creating users automatically creates access groups for them

Rockyy174:
From now on, creating a user automatically creates an access group for them. Previously, we were creating users in 5 different places in the application:

1) lib/operately/operations/company_adding.ex
2) lib/operately/operations/company_member_adding.ex
3) lib/operately/people/fetch_or_create_account_operation.ex
4) lib/mix/tasks/operately.add.account.to.company.ex
5) lib/operately/companies.ex

For the first 3, I've created the function `Operately.People.insert_person` which automatically inserts an access group for the user in the multi operation.

Although `Operately.People.create_person` is not being used anywhere in the application except for tests, I've also modified it so that it will create an access group for the user as well.

Number 4) was used to create users in the command line, but after https://github.com/operately/operately/pull/482 and https://github.com/operately/operately/pull/505, creating users using the command line is no longer necessary. So, it was decided that maintaining another way to create users was not necessary and, therefore, this file was deleted.

Number 5) was not being used anywhere in the application, so it was also deleted.


shiroyasha:
Looks good. Is it ready for merge? 🚀

Rockyy174:
Yes, we can merge it, @shiroyasha.

PR #648, opened 2024-06-27
Add data migration for creating access groups for users

Rockyy174:
I've created a data migration that creates an access group for every existing user.


Rockyy174:
I am aware that there are some old commits in this PR. As soon as https://github.com/operately/operately/pull/647 is merged, I will rebase it.

Rockyy174:
This one should be ready for merge as well.

PR #653, opened 2024-06-28
Creating person's access group also creates access group membership

Rockyy174:
I've updated `Operately.People.insert_person` so that, after creating an access group for the user, it also creates an access group membership. 

I've also made a data migration for creating an access group membership between existing users and their access groups.


lib/operately/access/access_group.ex
```diff
@@ -17,6 +17,18 @@ defmodule Operately.Access.Group do
   def changeset(group, attrs) do
     group
     |> cast(attrs, [:person_id, :company_id, :tag])
+    |> validate_one_association
```
shiroyasha, line 20: add `()`. It works without it as well, but it will fail a linter when we add it.

PR #674, opened 2024-07-03
Wordsmith Spaces form copy

markoa:
Before:
Screenshot 2024-07-03 at 11 28 10

After:
Screenshot 2024-07-03 at 11 53 10

Reasoning:
- Title: use active voice, this common in UI (will apply to other screens but don't want to do too much in my first PR of the kind)
- subtitle to describe the purpose of spaces - a little bit of help
- color / icon - less words is better
- Privacy / Internal - "Only" felt off
- Added an assuring sentence before the buttons that everything can be changed later so no worries.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -18,6 +18,7 @@ GEM
 
 PLATFORMS
   aarch64-linux
+  arm64-darwin-22
 
 DEPENDENCIES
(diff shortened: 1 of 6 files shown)


shiroyasha:
Looks good. 🚀

PR #706, opened 2024-07-05
feat: Add ProjectContributorRemoved operation

Rockyy174:
I've added `Operately.Operations.ProjectContributorRemoving` which deletes a contributor of a project and the access binding between the contributor's access group and the project's access context.

I've also moved the contributor removing process to the new API.

Previously, when deleting a contributor, we were also creating an `Update`. In the new `ProjectContributorRemoving` operation, we are creating an activity record instead.

Several tests were using `Operately.Projects.create_contributor` to create contributors which didn't create an access binding. So, when some of the tests attempted to delete the contributor, they would fail.
To fix that, I've delegated `Operately.Projects.create_contributor` to `Operately.Operations.ProjectContributorAddition`. Now, whenever a contributor is created, it consistently creates an access binding for them.


shiroyasha:
Looks good. One thing that we need to add is to display the new Operation in the feed. Should that be a new PR?

PR #767, opened 2024-07-15
fix: Make get_invitation and change_password_first_time public

Rockyy174:
When we migrated to the new API, `get_invitation` and `change_password_first_time` ended up as private operations. Therefore, inviting a new member was not working.


shiroyasha:
Good catch!

PR #768, opened 2024-07-15
feat: Add get_assignments query

Rockyy174:
I've added the `OperatelyWeb.Api.Queries.GetAssignments` query, which will be used in the new Review page to fetch all the due assignments for a specific person.

This query is a little different from the others and I found it very difficult to use our serializer to handle it. 

The thing is that we need to fetch information from many different tables, but just a few fields from each table. So, to use the serializer as it is, I would have to fetch way more data than we needed. 
Furthermore, the current `essential` and `full` levels for the structures I was working with didn't meet some of the requirements for the data, so there wasn't a straightforward way of using the serializer without extending it.

So, rather than fetching much more data than we needed and extending the serializer to handle a single use case, I decided to optimize the queries to fetch only the data we need and already in the required format.

### DIFF
--- a/test/operately_web/api/queries/get_assignments_test.exs
+++ b/test/operately_web/api/queries/get_assignments_test.exs
@@ -4,6 +4,7 @@ defmodule OperatelyWeb.Api.Queries.GetAssignmentsTest do
   import Operately.PeopleFixtures
   import Operately.ProjectsFixtures
   import Operately.GoalsFixtures
+  import Operately.UpdatesFixtures
 
   alias Operately.Repo
   alias Operately.Projects.Project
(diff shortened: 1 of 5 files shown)


test/operately_web/api/queries/get_assignments_test.exs
```diff
@@ -0,0 +1,267 @@
...
+      create_project(ctx, upcoming_date(), %{creator_id: another_person.id})
+
+      assert {200, %{assignments: assignments} = _res} = query(ctx.conn, :get_assignments, %{})
+
+      assert Repo.aggregate(Project, :count, :id) == 5
+      assert length(assignments) == 2
+
+      [p1, p2] = assignments
+
+      assert p1.id == today_project.id
```
shiroyasha, line 40: The API should return a short id version of project. You can generate one with OperatelyWeb.Paths.project_id().

PR #770, opened 2024-07-16
fix: search_people query encodes people IDs and mutations decode them

Rockyy174:
The mutations `update_project_contributor`, `edit_goal` and `create_project` weren't working because some of the IDs were not being decoded.

### DIFF
--- a/lib/operately_web/api/queries/search_people.ex
+++ b/lib/operately_web/api/queries/search_people.ex
@@ -1,5 +1,6 @@
 defmodule OperatelyWeb.Api.Queries.SearchPeople do
   use TurboConnect.Query
+  use OperatelyWeb.Api.Helpers
 
   import Ecto.Query, only: [from: 2, limit: 2]
 
@@ -20,7 +21,7 @@ defmodule OperatelyWeb.Api.Queries.SearchPeople do
   def call(conn, inputs) do
     Person
     |> match_by_full_name_or_title(inputs)
-    |> limit_to_company(conn.assigns.current_account.person.company_id)
+    |> limit_to_company((me(conn)).company_id)
     |> ignore_ids(inputs[:ignored_ids] || [])
     |> order_asc_by_match_position(inputs)
     |> exclude_suspended()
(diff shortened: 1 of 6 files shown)


lib/operately_web/api/queries/search_people.ex
```diff
@@ -59,15 +60,6 @@ defmodule OperatelyWeb.Api.Queries.SearchPeople do
   end
 
   def serialize(people) when is_list(people) do
-    %{people: Enum.map(people, &serialize/1)}
-  end
-
```
Rockyy174, line 64: `search_people` was not encoding the person's ID while `get_person` and `get_people` were. 
So, some of our mutations that were not decoding the person ID would sometimes fail and sometimes work, depending on how that person was loaded, which made it difficult for our tests to catch the problem.

For example, `create_goal` and `create_project` mutations were not decoding the IDs. However, in the frontend, the default champion for the project and goal is the logged in person, which is loaded using `get_person`, while the other options for champion and reviewer are loaded using `search_people`.

So, if a person other than the default was chosen as a champion, it would be a person with a non-ecoded ID and everything would work. 
And that's why our `project_creation` and `goal_creation` tests were all passing.

But if the default champion was selected, the mutations wouldn't work.

By changing `search_people` to also encode the IDs, our tests were able to find which mutations that weren't working.

lib/operately_web/api/queries/search_people.ex
```diff
@@ -59,15 +60,6 @@ defmodule OperatelyWeb.Api.Queries.SearchPeople do
...
-  def serialize(person) do
-    %{
-      id: person.id,
-      full_name: person.full_name,
-      title: person.title,
-      avatar_url: person.avatar_url
```
Rockyy174, line 70: `search_people` was not encoding the person's ID while `get_person` and `get_people` were. 
So, some of our mutations that were not decoding the person ID would sometimes fail and sometimes work, depending on how that person was loaded, which made it difficult for our tests to catch the problem.

For example, `create_goal` and `create_project` mutations were not decoding the IDs. However, in the frontend, the default champion for the project and goal is the logged in person, which is loaded using `get_person`, while the other options for champion and reviewer are loaded using `search_people`.

So, if a person other than the default was chosen as a champion, it would be a person with a non-ecoded ID and everything would work. 
And that's why our `project_creation` and `goal_creation` tests were all passing.

But if the default champion was selected, the mutations wouldn't work.

By changing `search_people` to also encode the IDs, our tests were able to find which mutations that weren't working.

shiroyasha:
Nice work! Thanks for fixing it 🙌

PR #772, opened 2024-07-17
feat: Review page

### LINKED ISSUE #592: Review screen
I'm responsible for reviewing many goals and I'm having a hard time catching up at the turn of the month.

My tools available are
- email inbox
- global goal hierarchy
- notifications

all three are very cluttered (notifications I now completely ignore tbh).

What I need is a clean list of things I **must do**, like status updates for goals I own + what I need to acknowledge.

Previously we had some discussions about how this plays with tasks but I think we need a simple zero to one solution first.

Rockyy174:
I've added the new Review page where champions and reviewers can easily see all their due assignments.

I'm marking this PR as a draft since it must be merged after https://github.com/operately/operately/pull/768 and  https://github.com/operately/operately/pull/771.

### DIFF
--- a/assets/js/layouts/DefaultLayout/Review.tsx
+++ b/assets/js/layouts/DefaultLayout/Review.tsx
@@ -1,7 +1,6 @@
-import React, { useCallback, useEffect, useState } from "react";
+import React from "react";
 
-import { getAssignmentsCount } from "@/api";
-import { useAssignmentsCount } from "@/api/socket";
+import { useAssignmentsCount } from "@/models/assignments";
 
 import { DivLink } from "@/components/Link";
 import { Paths } from "@/routes/paths";
@@ -9,7 +8,7 @@ import { IconCoffee } from "@tabler/icons-react";
 
 
 export function Review() {
-  const count = useCount();
+  const count = useAssignmentsCount();
 
   return (
     <DivLink
@@ -46,20 +45,4 @@ function AssignmentsCount({ count }: { count: number }) {
       {count}
     </div>
   );
-}
-
-function useCount() {
-  const [count, setCount] = useState(0);
-
-  const fetch = useCallback(() => {
-    getAssignmentsCount({})
-    .then((data) => {
-      setCount(data.count!);
-    })
-  }, []);
-
-  useEffect(() => fetch(), []);
-  useAssignmentsCount(fetch);
-
-  return count;
 }
--- a/assets/js/layouts/DefaultLayout/Review.tsx
+++ b/assets/js/layouts/DefaultLayout/Review.tsx
@@ -1,0 +1,65 @@
+import React, { useCallback, useEffect, useState } from "react";
+
+import { getAssignmentsCount } from "@/api";
+import { useAssignmentsCount } from "@/api/socket";
+
+import { DivLink } from "@/components/Link";
+import { Paths } from "@/routes/paths";
+import { IconCoffee } from "@tabler/icons-react";
+
+
+export function Review() {
+  const count = useCount();
+
+  return (
+    <DivLink
+      to={Paths.reviewPath()}
+      className="font-semibold flex items-center gap-1 cursor-pointer group hover:bg-base-accent px-1.5 py-0.5 rounded relative"
+      data-test-id="review-link"
+    >
+      <IconCoffee
+        size={20}
+        stroke={2}
+        className="mb-[3px]"
+      />
+      Review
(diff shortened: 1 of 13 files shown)


assets/js/layouts/DefaultLayout/Review.tsx
```diff
@@ -0,0 +1,65 @@
...
+        fontWeight: "900",
+      }}
+      data-test-id="review-link-count"
+    >
+      {count}
+    </div>
+  );
+}
+
+function useCount() {
```
shiroyasha, line 51: Let's create a model for assignments in `assets/js/models/assignments` and move this logic there. Models are usually a good mediator between the view part and the API, where we can store other data fetching and transformation logic.

As a rule of thumb, I would suggest avoiding the direct use of the api namespace in views directly, and almost always using the models as a middleman.

assets/js/pages/ReviewPage/page.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+import { IconTarget, IconHexagons } from "@tabler/icons-react";
+
+import { Paths } from "@/routes/paths";
+import { ReviewAssignment } from "@/api";
+import { useLoadedData } from "./loader";
+import { useNavigateTo } from "@/routes/useNavigateTo";
+
+
+type AssignmentType = "goal" | "project" | "goal_update" | "check_in";
```
shiroyasha, line 13: This is also a good candidate to be moved to the model.

assets/js/pages/ReviewPage/page.tsx
```diff
@@ -0,0 +1,168 @@
...
+}
+
+
+function formatDate(dateString: string) {
+  const date = new Date(dateString);
+  const options = { month: 'long', day: 'numeric' } as Intl.DateTimeFormatOptions;
+  return date.toLocaleDateString('en-US', options);
+}
+
+function calculateDaysAgo(dateString: string) {
```
shiroyasha, line 123: Move this to utils/time to centralize time related operations.

assets/js/pages/ReviewPage/page.tsx
```diff
@@ -0,0 +1,168 @@
...
+      )}
+    </div>
+  );
+}
+
+
+function formatDate(dateString: string) {
+  const date = new Date(dateString);
+  const options = { month: 'long', day: 'numeric' } as Intl.DateTimeFormatOptions;
+  return date.toLocaleDateString('en-US', options);
```
shiroyasha, line 120: There is a FormattedTime component, can we use it here?

assets/js/layouts/DefaultLayout/Review.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const [count, setCount] = useState(0);
```
shiroyasha, line 52: The `make gen` is also generating a hook for the query, called useGetAssignementCount that returns a refetch command as well. Can we use it here?

Pseudocode:

```
const { data, refetch } = useGetAssignmentCount();
const count = data.count;

useAssignmentsCount(() => refetch());
```

assets/js/layouts/DefaultLayout/Review.tsx
```diff
@@ -0,0 +1,65 @@
+import React, { useCallback, useEffect, useState } from "react";
+
+import { getAssignmentsCount } from "@/api";
+import { useAssignmentsCount } from "@/api/socket";
```
shiroyasha, line 4: Note for the some future development:

Events from the backend should have a name that resembles a signal or a notification, to make the code a bit more clear. Something like `useOnAssignmentCountChange` or `useAssignmentCountChangeSignal`. Afaik, the back-and implementation of sockets is not yet allowing this so we can't do it in this PR, but its a good food for thought in upcoming ones.

assets/js/layouts/DefaultLayout/Review.tsx
Rockyy174: Yes, that was right. I was able to simply the hook quite a bit by using `useGetAssignmentCount`. Thanks!

assets/js/pages/ReviewPage/page.tsx
Rockyy174: Yes, it was possible to use it. I didn't know about this component. Thank you!

Rockyy174:
I believe everything has been taken care of. Thank you for all the insights, @shiroyasha!

If there is anything else that can be improved, just let me know!

PR #773, opened 2024-07-19
feat: Add access level page for Spaces

Rockyy174:
I've added a page for managing the access level of spaces according to our wireframe.

On the new page, we can:

- Add new members
- Remove members
- Add the access level of individual members
- Add the access level for the space

Since now we can edit the access level on this page, I've deleted the page `SpaceEditPermissionsPage`.

The actions we do on the page `GroupMembersPage` can also be done on the new page, so we might also want to delete it in the future.


lib/operately_web/api/mutations/edit_space_members_permissions.ex
```diff
@@ -0,0 +1,32 @@
...
+  def call(conn, inputs) do
+    {:ok, group_id} = decode_id(inputs.group_id)
+
+    group = Operately.Groups.get_group!(group_id)
+    author = me(conn)
+    members = decode_ids(inputs.members)
+
+    {:ok, _} = Operately.Operations.GroupMembersPermissionsEditing.run(author, group, members)
+
+    {:ok, true}
```
shiroyasha, line 23: This should be `{:ok, %{success: true}}`.

Note for future: Some kind of linter or after call hook should be able to detect these kind of problems.

Rockyy174: Oh, I see. I will fix it right away.

shiroyasha:
Looks good! But it seems that we have some conflicts now that I have merged the other PRs.

Rockyy174:
It should be good now.

shiroyasha:
I see that there is a failed test for something related to groups. Is it a flaky one, or a real problem after rebase?

Rockyy174:
It seems it's a flaky one. 
I think I know what the problem is. Since it's a new test, I will change it before we merge it.

PR #774, opened 2024-07-22
feat: Add permissions filter to get_activities query

Rockyy174:
I've added a permissions filter `OperatelyWeb.Api.Queries.GetActivities` in order to load only the activities for which a person has view permissions.

I've added tests for several situations and everything seems to be working.

While I was testing this query, I found out that both `Operately.People.insert_person/2` and `Operately.People.create_person/1` are not creating an access group membership between the new person's access group and the company's access context.

Since `Operately.People.create_person/1` is used only for tests, I've already fixed it in this pull request.

I will fix `Operately.People.insert_person/2` in a separate pull request.

### DIFF
--- a/lib/operately_web/api/queries/get_activities.ex
+++ b/lib/operately_web/api/queries/get_activities.ex
@@ -6,6 +6,7 @@ defmodule OperatelyWeb.Api.Queries.GetActivities do
   alias Operately.Activities.Activity
   alias Operately.Activities.Preloader
   alias Operately.Companies.ShortId
+  alias Operately.Access.Binding
 
   import Ecto.Query, only: [from: 2, limit: 2, preload: 2]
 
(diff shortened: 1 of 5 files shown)


lib/operately_web/api/queries/get_activities.ex
```diff
@@ -98,4 +99,14 @@ defmodule OperatelyWeb.Api.Queries.GetActivities do
     from a in query, where: a.action in ^actions and a.action not in ^Activity.deprecated_actions()
   end
 
+  def filter_by_permissions(query, person) do
+    from(a in query,
+      join: c in assoc(a, :context),
+      join: b in assoc(c, :bindings),
+      join: g in assoc(b, :group),
+      join: m in assoc(g, :memberships),
+      where: m.person_id == ^person.id and b.access_level >= ^Binding.view_access(),
```
shiroyasha, line 108: Is this covering anonymous requesters as well?

Rockyy174: Interesting question. I didn't consider the possibility that this API would be public, but it makes sense. 
The thing is that we are using `company_id` which is extracted from the current logged in user:

```
company_id = conn.assigns.current_account.person.company_id
```

To enable this API to handle anonymous requests, we would have to include `company_id` to inputs.

Do you want me to handle that in this pull request or in a separate one?

shiroyasha: Ah, I see. The `company_id` will also be needed for probably every other API endpoint once we introduce the support for having multiple companies for a single account.

Let's not do that in this PR, I have some alternative strategy how to do it. I'll share it with you on Monday.

I'll merge this PR as is, and we can add that addition when we introduce the company_id for all api endpoints.

shiroyasha:
This one will need a rebase as well.

Rockyy174:
It should be good now.

PR #775, opened 2024-07-22
fix: People.insert_person/2 creates access membership with company group

### LINKED ISSUE #643: Zero state Company space says "Join this space"
This is after creating a self-hosted instance -- I should already be a member:

Screenshot 2024-06-26 at 13 30 40

Rockyy174:
I've updated `Operately.People.insert_person/2`. Now it also creates an access membership between the new person and the company's access group.

If we come across errors or bugs that suggest that there is no access membership between an existing person and the company's access group, running the data migration 015 should fix the problem.

### DIFF
--- a/lib/operately/companies.ex
+++ b/lib/operately/companies.ex
@@ -12,8 +12,8 @@ defmodule Operately.Companies do
 
   def list_companies(account = %Operately.People.Account{}) do
     Repo.all(
-      from c in Company, 
-        join: p in assoc(c, :people), 
+      from c in Company,
+        join: p in assoc(c, :people),
         where: p.account_id == ^account.id)
   end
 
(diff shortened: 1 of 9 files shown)


Rockyy174:
It turned out that the issue described in https://github.com/operately/operately/issues/643 was also because of `People.insert_person/2`.

The thing is that `People.insert_person/2` was not adding the new person as a member to the company's space. 
We are creating people in 3 different places in the application and we use `People.insert_person/2` in all of them, which means that a new person was never added as a member of the company's space. 

This issue has now been fixed everywhere.

Furthermore, `People.insert_person/2` is handling a lot of things and becoming quite long, so I've moved it into a separate module, `InsertPersonIntoOperation`, in order to keep the `People` context cleaner. 
The new module is delegated to `People.insert_person/2`, so the way we use it across the application didn't change.

PR #776, opened 2024-07-22
fix: login redirects to previously visited page

### LINKED ISSUE #728: Logging in doesn't redirect to where I was going
- got an email
- clicked 'view discussion'
- logged in (with google)
- got redirected to home, instead of the discussion

Rockyy174:
I've fixed the bug described in https://github.com/operately/operately/issues/728 by adding a `redirect_to` key to the session with the current path as value before the user is redirected to the login page. When the user logs in, if `redirect_to` is present in the session, `OperatelyWeb.AccountAuth.log_in_account/3` redirects them to that path.

I've tested it extensively on localhost and wrote some unit tests for it as well. 
Unfortunately, I couldn't test it with OAuth, but since it also uses `OperatelyWeb.AccountAuth.log_in_account/3`, it should be working, too.

### DIFF
--- a/lib/operately_web/account_auth.ex
+++ b/lib/operately_web/account_auth.ex
@@ -27,7 +27,7 @@ defmodule OperatelyWeb.AccountAuth do
   """
   def log_in_account(conn, account, params \\ %{}) do
     token = People.generate_account_session_token(account)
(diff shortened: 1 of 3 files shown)


shiroyasha:
Nice!

PR #781, opened 2024-07-23
feat: Project description can be collapsed and expanded

### LINKED ISSUE #689: Expand / Collapse long project descriptions
"Project overview should support longer text - Add the ↓ Expand ↓ or ↓ Read more ↓ if text is longer than certain amount of lines, or a character limit if it should stay short." -- Alex from Semaphore

Rockyy174:
Project descriptions can now be collapsed and expanded as suggested in https://github.com/operately/operately/issues/689.

Everything should be working, but we may want to improve the design before merging it. Here is how it looks:

Collapsed:

collapsed

Expanded:

expanded

Working with the rich text that we use with Tiptap is a bit tricky. I took a look at their documentation, but I didn't find any straightforward utils to use with it. So, I've added 2 rich text utils: `countCharacters` and `truncate`.

Also, I think that a good behavior for `truncate` would be to not truncate a person's name, so I wrote it this way. But we can easily change that.


markoa:
The second screenshot looks like it lost ability to format paragraphs etc?

Rockyy174:
I've just tested some formatting and it should still be working, but I probably didn't test all the options. 

I've added some unit tests as well, but if you think it's necessary, I can add some more tests to make sure everything is working.

Collapsed:
collapsed

Expanded:
expanded

markoa:
Looks good, but I'll leave the code review to @shiroyasha for this one.

shiroyasha:
Looks good! 👍 

There is a similar rich text utility for truncating in `components/RichContent`. That one is used for summarizing for the content for the feed.

Can you move this implementation from `utils/richText` to that folder so that we keep all the rich content transformations in one place?

Rockyy174:
I've just moved the files to `components/RichContent`. Let me know if any other changes are needed!

PR #783, opened 2024-07-24
fix: Project list avatars no longer flow out of container

### LINKED ISSUE #456: Project List - Avatar list is flowing out of the container
avatar-list

Rockyy174:
I've fixed the issue described in https://github.com/operately/operately/issues/456. 

The problem was that the "next-milestone-info" container had a hard-coded width. Removing it was enough to fix it.
I've also added some margin-bottom to the avatars so that they wouldn't flow over the text.

Before:
before

Now:
after


markoa (approved):
feel free to merge

PR #785, opened 2024-07-24
fix: Suspended people excluded from space and project queries

### LINKED ISSUE #651: Suspended organization members are listed on projects

Rockyy174:
I've fixed the issue described in https://github.com/operately/operately/issues/651.

Now both project and space queries exclude members who have been suspended.


shiroyasha:
Nice! I like that we have solved on the schema level.

PR #797, opened 2024-07-26
fix: Adding and editing tasks with assignees were broken

### LINKED ISSUE #796: Assigning a person to a task not working
It seems that some of our Tasks APIs are not working. My guess is that it's all related to the new short IDs.

create_task:
cant-create-task

update_task:
cant-create-task-3

search_people when trying to add more than one person:
cant-create-task-2

Rockyy174:
I've fixed the issue described in https://github.com/operately/operately/issues/796.

We didn't realize these APIs were broken earlier because we didn't have feature tests for Tasks, so I've decided to add them in this pull request.


shiroyasha:
Nice! Feature tests for tasks were definitely needed.

PR #798, opened 2024-07-26
fix: Add margin-top to "Join this Space" button

Rockyy174:
I've noticed that there was no space between the "Join this Space" and the "Manage Access" buttons.

Previously:
bug

Now:
bug-fixed


shiroyasha:
👍 It is better than previously, but I'm noting it down that we need to revisit this page to make visual adjustments. Having two buttons on a screen with completely different design style is 🙅

PR #803, opened 2024-07-29
feat: Add activities and notifications for adding and removing space members

### LINKED ISSUE #671: Operations that add and remove members from and to spaces not recording activity
The following operations are not creating an activity record:

- lib/operately/operations/group_members_adding.ex
- lib/operately/operations/group_member_removing.ex

Rockyy174:
I've handled the issue described in https://github.com/operately/operately/issues/671.

Now adding members to a space or removing them from it creates an activity record. These records are going to be displayed in the Feed.
Also, members who are added to a space should receive an email notification.

### DIFF
--- a/lib/operately_email/templates/space_members_added.text.eex
+++ b/lib/operately_email/templates/space_members_added.text.eex
@@ -1,0 +1,3 @@
+<%= short_name(@author) %> added yout to the <%= @space.name %> space.
+
+Link: <%= @link %>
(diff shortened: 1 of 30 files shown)


lib/operately_email/templates/space_members_added.text.eex
```diff
@@ -0,0 +1,3 @@
+<%= short_name(@author) %> added yout to the <%= @space.name %> space.
```
shiroyasha, line 1: Typo `yout` -> `you`

shiroyasha:
Looks good 👍 Apart from the small typo, everything else is ready for 🚢

lib/operately_email/templates/space_members_added.text.eex
Rockyy174: 😅

shiroyasha:
It seems that the test failed, or is it flaky?

Rockyy174:
I think it's flaky because it had passed before I fixed the typo. But is's failed about 3 times already 😬

PR #851, opened 2024-08-06
feat: Add permissions filter to GetPeople query

Rockyy174:
I've added a permissions filter to the `OperatelyWeb.Api.Queries.GetPeople` query.

I noticed that the `filter_by_view_access` function was allowing suspended people to use the API. This issue is also handled in this PR.

### DIFF
--- a/lib/operately/access/filters.ex
+++ b/lib/operately/access/filters.ex
@@ -14,7 +14,9 @@ defmodule Operately.Access.Filters do
       join: b in assoc(c, :bindings),
       join: g in assoc(b, :group),
       join: m in assoc(g, :memberships),
+      join: p in assoc(m, :person),
       where: m.person_id == ^person_id and b.access_level >= ^access_level,
+      where: is_nil(p.suspended_at),
       distinct: true
     )
   end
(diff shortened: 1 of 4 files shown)


lib/operately/access/filters.ex
```diff
@@ -14,7 +14,9 @@ defmodule Operately.Access.Filters do
       join: b in assoc(c, :bindings),
       join: g in assoc(b, :group),
```
Rockyy174, line 15: It's worth mentioning that, since I excluded suspended people at the schema level, line 17 was enough. 
However, I decided to add line 19 to make it explicit, otherwise in the future we may not remember why we need the extra join in line 17.

lib/operately/access/filters.ex
```diff
@@ -14,7 +14,9 @@ defmodule Operately.Access.Filters do
...
       join: m in assoc(g, :memberships),
+      join: p in assoc(m, :person),
       where: m.person_id == ^person_id and b.access_level >= ^access_level,
+      where: is_nil(p.suspended_at),
```
Rockyy174, line 19: It's worth mentioning that, since I excluded suspended people at the schema level, line 17 was enough. 
However, I decided to add line 19 to make it explicit, otherwise in the future we may not remember why we need the extra join in line 17.

PR #872, opened 2024-08-08
feat: Add permissions filter to ArchiveProject mutation

Rockyy174:
I've added a permissions filter to the `OperatelyWeb.Api.Mutations.ArchiveProject` mutation.

### DIFF
--- a/test/operately_web/api/mutations/archive_project_test.exs
+++ b/test/operately_web/api/mutations/archive_project_test.exs
@@ -24,7 +24,7 @@ defmodule OperatelyWeb.Api.Mutations.ArchiveProjectTest do
       Map.merge(ctx, %{creator: creator, creator_id: creator.id, space_id: space.id})
     end
 
-    test "company members CAN'T see resource", ctx do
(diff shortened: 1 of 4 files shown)


test/operately_web/api/mutations/archive_project_test.exs
```diff
@@ -1,3 +1,228 @@
...
+
+    test "company members CAN'T perfrom action", ctx do
+      project = create_project(ctx, company_access_level: Binding.edit_access())
+
+      assert {403, %{message: message}} = mutation(ctx.conn, :archive_project, %{project_id: Paths.project_id(project)})
+      assert message == "You don't have permission to perform this action"
+      refute_project_archived(project)
+    end
+
+    test "company members have access", ctx do
```
shiroyasha, line 43: It would be easier to understand if the name test would follow an "assumption -> outcome" pattern. e.g. "company members who have access can archive a project".

Rockyy174: Yes, it makes more sense like this!

Issue #813, opened 2024-07-31
Docs & Files in projects


markoa:
Assuming we do this for spaces first (text documents, file uploads, folders, links). We would then carry it over to projects.

Rockyy174:
Already implemented.

Issue #598, opened 2024-06-18
Discussions in projects


markoa:
Obvious requirement.

markoa:
To add, from a "production user's perspective", it's a pretty big issue that I can't initiate structured communication about an ongoing project inside a company.

markoa:
Removing `spec-needed` from this one because we have the needed design elements for the core page in #882.

PR #882, opened 2024-08-09
Mock: Project page message board, resources, improved milestones

markoa:
This branch is for reference, not to be merged.

Contains a modification of the project page to serve as a mockup for #598 and #813.

what's new/different:
- added message board
- resources as a mix of files, docs, links
- milestones redesigned
- using (x) numbers next to things to communicate how alive/large a project is and eliminate "View all" links
- replaced zero state copy from neutral/negative to positive/descriptive (eg. "No milestones defined yet." → "Outline your project timeline with milestones.")
- replaced secondary buttons with those from Tailwind UI package. this is a tiny step in the direction of reusing well developed components from an established system instead of reinventing the wheel.

image

### DIFF
--- a/assets/js/pages/ProjectPage/NextMilestone.tsx
+++ b/assets/js/pages/ProjectPage/NextMilestone.tsx
@@ -30,14 +30,12 @@ function MilestonesZeroState({ project }) {
   const editPath = Paths.projectEditTimelinePath(project.id!);
 
   const editLink = (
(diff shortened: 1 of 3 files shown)


shiroyasha:
To check if we have this covered:

When the project has a description (not in zero-state), how do you edit it?

markoa:
I'd keep it as it is rn. 
Screenshot 2024-09-04 at 13 42 56

PR #924, opened 2024-08-15
feat: Change zero-state microcopy on project page.

markoa:
From "There is no X" to "Here's what you can do".

Backported from #882 except for the goal linking which is a new idea.

### DIFF
--- a/assets/js/pages/ProjectPage/Header.tsx
+++ b/assets/js/pages/ProjectPage/Header.tsx
@@ -67,11 +67,12 @@ function ParentGoal({ project }: { project: Projects.Project }) {
   } else {
     content = (
       <div className="text-sm text-content-dimmed mx-1 font-medium">
-        <Link to={Paths.editProjectGoalPath(project.id)} className="underline hover:text-link-hover">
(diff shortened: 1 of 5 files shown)


assets/js/pages/ProjectPage/Header.tsx
```diff
@@ -65,7 +65,7 @@ function ParentGoal({ project }: { project: Projects.Project }) {
       </>
     );
   } else {
-    content = <div className="text-sm text-content-dimmed mx-1 font-medium">Not yet connected to a goal</div>;
+    content = <div className="text-sm text-content-dimmed mx-1 font-medium"><a href={Paths.editProjectGoalPath(project.id)} className="underline hover:text-link-hover">Link this project to a goal</a> for context and purpose</div>;
```
shiroyasha, line 68: Please use `<Link ...` from components/Link. Two reasons:

1. `<a>` kind of links are not aware of React.Router and skip over the loading logic.
2. Let's use a consistent style for links across the app, so if we need, we can change it in one spot.

markoa: Roger that, pushed an update.

PR #940, opened 2024-08-16
feat: Add permissions filter to EditGroup and EditSpacePermissions

Rockyy174:
I've added a permissions filter to the `OperatelyWeb.Api.Mutations.EditGroup` and `OperatelyWeb.Api.Mutations.EditSpacePermissions` mutations.

### DIFF
--- a/lib/operately_web/api/mutations/edit_group.ex
+++ b/lib/operately_web/api/mutations/edit_group.ex
@@ -17,16 +17,16 @@ defmodule OperatelyWeb.Api.Mutations.EditGroup do
 
   def call(conn, inputs) do
     person = me(conn)
+    {:ok, space_id} = decode_id(inputs.id)
 
-    with {:ok, space_id} <- decode_id(inputs.id),
-        {:ok, space, access_level} <- Groups.get_group_and_access_level(space_id, person.id),
-        true <- Permissions.can_edit(access_level)
-    do
-      execute(person, space, inputs)
-    else
+    case Groups.get_group_and_access_level(space_id, person.id) do
+      {:ok, space, access_level} ->
+        if Permissions.can_edit(access_level) do
+          execute(person, space, inputs)
+        else
+          {:error, :forbidden}
+        end
       nil -> {:error, :not_found}
-      false -> {:error, :forbidden}
-      _ -> {:error, :bad_request}
     end
   end
 
--- a/lib/operately_web/api/mutations/edit_group.ex
+++ b/lib/operately_web/api/mutations/edit_group.ex
@@ -26,7 +26,7 @@ defmodule OperatelyWeb.Api.Mutations.EditGroup do
         else
           {:error, :forbidden}
         end
-      nil -> {:error, :not_found}
+      {:error, reason} -> {:error, reason}
     end
   end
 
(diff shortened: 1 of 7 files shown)


Rockyy174:
I've implemented two variants of the mutation handler: 
1) One using `case` and `if` statements
2) Another using a railway approach

To be honest, I'm not sure which one I prefer. What do you think, @shiroyasha?

I will also implement a 3) one using `with` statement, but I will have to modify the `Filters` and `Permissions` module, so I will submit it separately as draft.

lib/operately_web/api/mutations/edit_group.ex
```diff
@@ -13,14 +16,22 @@ defmodule OperatelyWeb.Api.Mutations.EditGroup do
...
 
   def call(conn, inputs) do
+    person = me(conn)
     {:ok, space_id} = decode_id(inputs.id)
-    space = Operately.Groups.get_group!(space_id)
 
-    {:ok, space} = Operately.Groups.edit_group_name_and_purpose(me(conn), space, %{
-      name: inputs.name,
-      mission: inputs.mission
-    })
```
Rockyy174, line 22: Using `case` and `if`

lib/operately_web/api/mutations/edit_group.ex
```diff
@@ -13,14 +16,22 @@ defmodule OperatelyWeb.Api.Mutations.EditGroup do
...
+    case Groups.get_group_and_access_level(space_id, person.id) do
+      {:ok, space, access_level} ->
+        if Permissions.can_edit(access_level) do
+          execute(person, space, inputs)
```
Rockyy174, line 25: Using `case` and `if`

lib/operately_web/api/mutations/edit_space_permissions.ex
```diff
@@ -12,12 +15,34 @@ defmodule OperatelyWeb.Api.Mutations.EditSpacePermissions do
   end
 
   def call(conn, inputs) do
-    {:ok, id} = decode_id(inputs.space_id)
+    person = me(conn)
+
+    load_group(inputs, person)
+    |> verify_permissions()
+    |> execute(person, inputs)
+  end
```
Rockyy174, line 23: Second approach

lib/operately_web/api/mutations/edit_group.ex
shiroyasha: This one looks quite clean, but if a function is returning {:ok, space, access_level} i would expect the alternative answer to be {:error, reason}.

For example:

```
case Groups.get_group_and_access_level(space_id, person.id) do
  {:ok, space, access_level} ->
  {:error, :not_found} -> ...
end
```

lib/operately_web/api/mutations/edit_space_permissions.ex
shiroyasha: This one is nice, but the fact that we need to implement multiple alternative function is PITA.

lib/operately_web/api/mutations/edit_group.ex
Rockyy174: Makes sense

Rockyy174:
I've updated both EditGroup and EditSpacePermissions to use the solution with the `case` and `if` statements. Let me know if you have any other improvements to suggest!

shiroyasha:
Looks great! Nothing else for changing.

PR #942, opened 2024-08-17
feat: Add permissions filter to EditGroup and EditSpacePermissions

Rockyy174:
As mentioned in https://github.com/operately/operately/pull/940, this is another approach for the mutations handler where we use a `with` statement.

In my opinion, this is the cleanest variant and my favorite. 

A drawback in this approach is the fact that we want to use the same functions in `Operately.Groups.Permissions` for the mutations/queries and for the frontend. However, in the `with` statement we need ok/error tuples, while in the frontend we need booleans. 

So, we need more boilerplate to achieve the final result:

```
  def calculate_permissions(access_level) do
    %__MODULE__{
      can_view: can_view(access_level, boolean: true),
      can_edit: can_edit(access_level, boolean: true),
      can_edit_permissions: can_edit_permissions(access_level, boolean: true),
    }
  end

  def can_view(access_level, boolean: true), do: access_level >= Binding.view_access()
  def can_view(access_level), do: ok_or_error(access_level >= Binding.view_access())

  def can_edit(access_level, boolean: true),do: access_level >= Binding.edit_access()
  def can_edit(access_level),do: ok_or_error(access_level >= Binding.edit_access())

  def can_edit_permissions(access_level, boolean: true), do: access_level >= Binding.full_access()
  def can_edit_permissions(access_level), do: ok_or_error(access_level >= Binding.full_access())

  defp ok_or_error(true), do: {:ok, :allowed}
  defp ok_or_error(false), do: {:error, :forbidden}
```

Instead of simply:

```
  def calculate_permissions(access_level) do
    %__MODULE__{
      can_view: can_view(access_level),
      can_edit: can_edit(access_level),
      can_edit_permissions: can_edit_permissions(access_level),
    }
  end

  def can_view(access_level), do: access_level >= Binding.view_access()
  def can_edit(access_level), do: access_level >= Binding.edit_access()
  def can_edit_permissions(access_level), do: access_level >= Binding.full_access()
```

What do you think, @shiroyasha?

### DIFF
--- a/lib/operately/access/filters.ex
+++ b/lib/operately/access/filters.ex
@@ -14,9 +14,13 @@ defmodule Operately.Access.Filters do
       where: m.person_id == ^person_id and b.access_level >= ^Binding.view_access(),
       where: is_nil(p.suspended_at),
       group_by: r.id,
-      select: {:ok, r, max(b.access_level)}
+      select: {r, max(b.access_level)}
     )
     |> Repo.one()
+    |> case do
+      nil -> {:error, :not_found}
+      {r, level} -> {:ok, r, level}
+    end
   end
 
   def filter_by_view_access(query, person_id, opts \\ []) do
--- a/lib/operately/access/filters.ex
+++ b/lib/operately/access/filters.ex
@@ -4,6 +4,21 @@ defmodule Operately.Access.Filters do
   alias Operately.Repo
   alias Operately.Access.Binding
 
+  def get_resource_and_access_level(query, person_id) do
+    from([resource: r] in query,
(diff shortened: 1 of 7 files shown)


shiroyasha:
Approach (1) is cleaner imho. This one adds complexity to an otherwise simple calculation, that is only needed in some unrelated location in the codebase.

Rockyy174:
I liked this approach, but I'm probably too biased, since I wrote it 😅  
So, let's stick to approach 1

shiroyasha:
Closing this as we have merged https://github.com/operately/operately/pull/940

PR #956, opened 2024-08-20
chore: Add buttons to design system.

markoa:
This is an attempt to contribute a little to streamlining our UI elements.

These are basically buttons from Tailwind UI with primary skinned to our `accent-1`.


markoa:
There are probably more sizes here than we need: 

Screenshot 2024-08-20 at 18 19 44

But it doesn't seem appropriate to cover in this PR. After putting these to work on several screens it should become obvious which ones are not needed.

shiroyasha:
Just a sanity check, do they work with dark mode?

PR #974, opened 2024-08-21
feat: New button in top navigation.

markoa:
This button opens a drop down menu with all things you can create in Operately.

This immediately solves the issue of making inviting new people more obvious. But it just makes things easier overall.

Screenshot 2024-08-21 at 13 55 04

---

I copy-pasted from `CompanyDropdown`. Not sure if there's room to DRY.


shiroyasha:
Looks great 👌 

Some DRY-ing can be done, but it is not essential before we merge the PR.

PR #1002, opened 2024-08-26
feat: Add SubscribersSelector component

Rockyy174:
I've added the component `SubscribersSelector` and the context `SubscriptionsContext`.

I tried to follow the wireframe as strictly as possible, but most likely we still need some adjustments in the design.

Also, once I add and update the API endpoints, other adjustments in the functionality will be necessary as well, such as displaying the Space name or whether it's a Space, Goal, Project, etc.
But overall the component and context should be mostly ready.

https://github.com/user-attachments/assets/a275adb5-bbe9-4498-a059-1b787adae512

### DIFF
--- a/assets/js/features/Subscriptions/SubscribersSelector.tsx
+++ b/assets/js/features/Subscriptions/SubscribersSelector.tsx
@@ -72,7 +72,7 @@ function SelectedPeople({ people, hide }) {
   if (people.length === 0 || hide) return <></>;
 
   return (
-    <div className="flex gap-2 ml-6">
+    <div className="flex flex-wrap gap-2 ml-6">
       {people.map((person) => (
         <Avatar person={person} size="tiny" key={person.id} />
       ))}
--- a/assets/js/features/Subscriptions/SubscribersSelector.tsx
+++ b/assets/js/features/Subscriptions/SubscribersSelector.tsx
@@ -115,7 +115,7 @@ function SelectorModal({ showSelector, setShowSelector }: SelectorModalProps) {
       minHeight="200px"
     >
       <Forms.Form form={form}>
-        <div className="flex flex-col gap-6">
+        <div className="flex flex-col gap-6 max-h-[400px] overflow-y-auto">
           <div className="flex items-center gap-3">
             <ActionLink onClick={handleSelectEveryone}>Select everyone</ActionLink>
             <span className="text-content-dimmed">&middot;</span>
--- a/assets/js/features/Subscriptions/SubscribersSelector.tsx
+++ b/assets/js/features/Subscriptions/SubscribersSelector.tsx
@@ -42,11 +42,7 @@ export function SubscribersSelector() {
           onClick={() => setShowSelector(true)}
         />
 
-        <SubscriptionOption
-          label="No one"
-          value={Options.NONE}
-          people={[]}
-        />
+        <SubscriptionOption label="No one" value={Options.NONE} people={[]} />
       </RadioGroup>
 
       <SelectorModal showSelector={showSelector} setShowSelector={setShowSelector} />
--- a/assets/js/features/Subscriptions/SubscribersSelector.tsx
+++ b/assets/js/features/Subscriptions/SubscribersSelector.tsx
@@ -1,0 +1,143 @@
+import React, { useState, useMemo } from "react";
+
+import { useSubscriptionsContext, Options } from "./SubscriptionsContext";
+import Forms from "@/components/Forms";
+import Modal from "@/components/Modal";
+import Avatar from "@/components/Avatar";
+import { Radio, RadioGroup } from "@/components/Form";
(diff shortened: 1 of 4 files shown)


shiroyasha:
Assuming that the we will implement the "All (N) people who are members of the space" with the avatar list will be implemented later, I have the following visual changes to request:

- The separator between Select everyone and Select no one should be a &middot; element with text-content-dimmed color.
- Change "Who should be notified?" -> "When I post this, notify:"
- Reduce the prominence of the separator between the people in the modal should be `bg-stroke-subtle`.
- The modal title is in a weird place. Can it be in the top-left corner?
- The modal is displayed quite low on the screen, I fear that it will look weird on smaller screens. Can you reduce the distance from the top? If the modal is a separate component, we can do it in another PR to avoid too many changes in this PR.

assets/js/features/Subscriptions/SubscribersSelector.tsx
```diff
@@ -0,0 +1,160 @@
...
+}
+
+
+function ActionText({text, onClick}) {
+  const isDark = useIsDarkMode();
+
+  return (
+    <span
+      className={classNames(
+        isDark ? "text-blue-300" : "text-blue-700",
```
shiroyasha, line 152: Let's not introduce new colors and follow the ones present in the localhost:4000:/__design__. Ideally, this component would live in `compoents/Links` nicknamed as ActionLink and followed the design and usage pattern of other links. 

Maybe do this in another PR, and when it is ready, rebase from main here.

assets/js/features/Subscriptions/SubscriptionsContext.tsx
```diff
@@ -0,0 +1,61 @@
+import React from "react";
```
shiroyasha, line 1: It feels like this file is not formatted with our default formatter for typescript defined in package.json, or at least Vim is generating a different formatting when I save the file.

Can you check if you have auto-formatting enabled?

I would like to follow the logic from the golang team https://www.youtube.com/watch?v=PAAkCSZUG1c&t=523s.

assets/js/features/Subscriptions/SubscribersSelector.tsx
```diff
@@ -0,0 +1,160 @@
...
+          <span className="text-[7px] content-accent">&#9830;</span>
+          <ActionText
+            text="Select no one"
+            onClick={handleSelectNoone}
+          />
+        </div>
+
+        <div>
+          {people.map(person => (
+            <div key={person.id} className="flex gap-4 border-b border-content-subtle px-2 pb-4 mb-4 last:border-0 last:mb-0">
```
shiroyasha, line 105: Extract the items as a seperate function to reduce the complexity of this function.

assets/js/features/Subscriptions/SubscribersSelector.tsx
```diff
@@ -0,0 +1,160 @@
...
+  const saveSelection = (ids: string[]) => {
+    setSelectedPeople(people.filter(person => ids.includes(person.id!)));
+    setShowSelector(false);
+  }
+
+  return (
+    <div>
+      <p className="text-lg font-bold mb-2">Who should be notified?</p>
+
+      <RadioGroup name="privacy-level" onChange={setSubscriptionType} defaultValue={subscriptionType} >
```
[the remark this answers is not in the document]
Rockyy174: :dizzy_face:

assets/js/features/Subscriptions/SubscribersSelector.tsx
Rockyy174: PR with ActionLink: https://github.com/operately/operately/pull/1006

shiroyasha:
Functionality wise looks good. 👍 We might make some light UI tweaks, but we can do that in another PR.

One thing that could be a blocker is: What happens if you have more people in the space, e.g. 20 people. How does the modal behave?

Rockyy174:
After 80vh max-height was added to the Modal, if the content doesn't fit, you have to scroll down to see it.


https://github.com/user-attachments/assets/85cd2cde-1aed-4531-a8ad-0c6ca58e1bea

Rockyy174:
As for the little icon list, I will add "flex-wrap" to make sure the icons don't flow out of the container.

shiroyasha:
In the modal, let's make the save button always visible, and the people list scrollable.
 
Reasoning: When the buttons are not visible, users might assume that you can simply select a person from the top and click on X.

Rockyy174:
Here is how it looks after adding a max-height of 400px to the modal inner content:


https://github.com/user-attachments/assets/0c893341-a1db-4e99-999c-caa293b5d2d8

shiroyasha:
I think everything is ready. 

I would merge it, but the tests are complaining that this component is not used anywhere, which a good check to have to avoid dead code.

Let's wait for the implementation on the project check-ins before we merge it.

Issue #969, opened 2024-08-21
text-red-500 should be renamed to text-error-base


shiroyasha:
The usage of direct color names (like text-red-500) is not following our advice from the design page.

PR #1004, opened 2024-08-27
chore: Alerts in design system.

markoa:
We're going to need this in a UI change coming soon.

Screenshot 2024-08-27 at 11 36 18

### DIFF
--- a/assets/js/components/Alerts/index.tsx
+++ b/assets/js/components/Alerts/index.tsx
@@ -1,0 +1,75 @@
+import React from "react";
+import { TestableElement } from "@/utils/testid";
+
+import { IconAlertTriangleFilled, IconCircleXFilled, IconInfoCircleFilled } from "@tabler/icons-react";
+
+interface Props extends TestableElement {
+  message: string | JSX.Element;
+  description?: string | JSX.Element;
+}
+
+export function InfoAlert(props: Props) {
+  return (
+    <div data-test-id={props.testId} className="rounded-md bg-blue-50 p-4">
+      <div className="flex">
+        <div className="flex-shrink-0">
+          <IconInfoCircleFilled aria-hidden="true" className="h-5 w-5 text-blue-400" />
+        </div>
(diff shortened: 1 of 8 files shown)


assets/js/components/Alerts/index.tsx
```diff
@@ -0,0 +1,69 @@
+import React from "react";
+
+import { IconAlertTriangleFilled, IconCircleXFilled, IconInfoCircleFilled } from "@tabler/icons-react";
+
+export function InfoAlert({ message, description }: { message: string | JSX.Element, description?: string | JSX.Element }) {
```
shiroyasha, line 5: The typing is identical for all three functions. Please extract with:

```
interface Props {
  message: string | JSX.Element;
  description: string | JSX.Element;
}
```

```
export function InfoAlert(props: Props) {
}
```


And then, extend it with a the TestableElement trait:

```
import { TestableElement } from "@/utils/testid";

interface Props extends TestableElement {
  message: string | JSX.Element;
  description: string | JSX.Element;
}
```

```
export function InfoAlert(props: Props) {
  return <div data-test-id={props.testId}>...
}
```

shiroyasha:
The other thing I need to ask you is to test it out in dark mode, to make sure that all the elements we have in the design are working in both environments.

This was so far not possible on the design page, but I've just merged a PR that allows it https://github.com/operately/operately/pull/1005. You will need to rebase from main to test it.

markoa:
@shiroyasha Thanks for the reminder about dark mode. Made it work:
Screenshot 2024-08-27 at 18 02 53
Screenshot 2024-08-27 at 18 02 59

I renamed them to callouts (inspired by Radix) because these widgets will not only be used as form alerts but also deeper on pages.

ℹ️  Two issues remain that I would tackle separately: #969 and secondary buttons not working well in dark mode.

shiroyasha:
Looks good. Thanks for updating it for the darkness.

PR #1008, opened 2024-08-27
feat: Add MultiPeopleSelectField form field

Rockyy174:
I've added a new form field: `MultiPeopleSelectField`

### DIFF
--- a/assets/js/components/Forms/FormState.tsx
+++ b/assets/js/components/Forms/FormState.tsx
@@ -1,3 +1,5 @@
+import { Dispatch, SetStateAction } from "react";
+
 type State = "idle" | "submitting";
 
 export type MapOfFields = Record<string, Field<any>>;
@@ -15,8 +17,8 @@ export interface FormState<FieldTypes extends MapOfFields> {
 
 export type Field<T> = {
   value: T | null | undefined;
-  setValue: (value: T | null | undefined) => void;
-  initial: T | null | undefined;
+  setValue: Dispatch<SetStateAction<T | undefined>>;
+  initial?: T | null | undefined;
   optional?: boolean;
   validate: () => string | null;
 };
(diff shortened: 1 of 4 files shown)


assets/js/components/Forms/MultiPeopleSelectField.tsx
```diff
@@ -0,0 +1,49 @@
+import React from "react";
+
+import Avatar from "../Avatar";
```
shiroyasha, line 3: Use absolute path to make it easier to move the file around if necessary `@/components/Avatar.`

assets/js/components/Forms/FormState.tsx
```diff
@@ -15,8 +17,8 @@ export interface FormState<FieldTypes extends MapOfFields> {
 
 export type Field<T> = {
   value: T | null | undefined;
-  setValue: (value: T | null | undefined) => void;
-  initial: T | null | undefined;
+  setValue: Dispatch<SetStateAction<T | undefined>>;
+  initial?: T | null | undefined;
```
shiroyasha, line 21: My TS typefoo is a bit weak. What is the difference between `initial?: ` and `initial: ... | undefined` and `initial?: ... | undefined`.

shiroyasha:
Everything looks good. I've added some nitpick comments.

assets/js/components/Forms/FormState.tsx
Rockyy174: Good question. When I added `?`, I didn't notice that `undefined` was already there. 

I also thought that `?` and `undefined` were the same thing, but without `?`, TS complains.
It seems there is a difference.

Rockyy174:
I've updated the import and double checked `type Field<T>` (it won't work without `?`).
This PR should be ready then.

assets/js/components/Forms/FormState.tsx
shiroyasha: We are still young typescript jedies, it seems that both you and me don't yet fully understand every typescript typing option 👦🏻

PR #1011, opened 2024-08-27
chore: Replace text-red-500 with a semantic color name

shiroyasha:
fixes: https://github.com/operately/operately/issues/969

I've also separated the colors in three categories. To have a good judgment about text colors I need to see actual text on the page.

The currently nicknamed "Miscellaneous" will most likely become UI elements, or Form colors.

Screenshot 2024-08-27 at 18 31 18

### DIFF
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -32,6 +32,7 @@ body[data-theme="light"] {
   --color-content-accent: rgba(0, 0, 0, 1);
   --color-content-dimmed: rgba(0, 0, 0, 0.6);
   --color-content-subtle: rgba(0, 0, 0, 0.3);
+  --color-content-error: var(--color-red-500);
 
   --color-stroke-base: rgba(0, 0, 0, 0.1);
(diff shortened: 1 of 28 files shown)


markoa:
I think that most if not all these uses that appear in the diff could be replaced with an error callout component.

markoa:
Re. color categorization, let's see if we can take inspiration from somewhere so we avoid words like Miscellaneous etc.

Eg. [Radix](https://www.radix-ui.com/colors/custom) categorizes colors as
- background
- interactive components (ie. background for forms, callouts, etc)
- borders and separators
- solid colors (buttons, solid backgrounds like avatars)
- text (secondary and primary)

shiroyasha:
Good names, thanks for sharing.

Regarding this:

> I think that most if not all these uses that appear in the diff could be replaced with an error callout component.

Perhaps, but not in this PR. Most of them are coming from some form elements where the first action item is adopt the components/Forms, which eliminates most of these instances.

PR #1016, opened 2024-08-28
chore: Success callout

markoa:
also removed some leftover unused color definitions.


shiroyasha:
Looks good. Merging and will fix the problem with the newly introduced fixer on main.

PR #1019, opened 2024-08-28
feat: Creating a check-in also creates a subscription list

Rockyy174:
I've updated the create-check-in form so that, when someone checks in a project, they can also choose who will be notified. 

The `PostProjectCheckIn` mutation and `ProjectCheckIn` operation take the new data and create a subscription list for the check-in and subscriptions for all the people who were selected in the form.

So, from now on, all check-ins will have a subscription list. In the upcoming PRs, I will add a data migration for creating subscription lists for existing check-ins. Then, I will update the check-in notifications to start using the new subscription lists.

I'm marking this PR as a draft because I'm not so sure about how to handle some details/edge cases:

- As options for "who will be notified", I'm displaying only the project contributors. Is that correct? Or should I display everyone in the Space or Company instead?

- In the current solution, the Reviewer may not be included in the subscription list and, therefore, not be notified. That doesn't seem right. If my assumption is correct, I can think of some solutions:

    1) Should the Reviewer always be selected in the form and the user not able to unselect? (form level) 
    2) Should the backend always add the Reviewer to the list regardless of the form?  (operation level)
    3) Should the Reviewer always be notified regardless of the list? (notification level)

- Currently, if there is only one person to be notified, the text will be: "All 1 people who are contributors of...". 
I could add a condition to display "1 person who is contributor of...", but it still sounds a bit weird. Is there a more appropriate text I could use? 🤔 

https://github.com/user-attachments/assets/f66ad300-1aee-4620-b8a4-3e73c6b06241


shiroyasha:
cc-ing @markoa for the copy questions.

assets/js/features/Subscriptions/utils.tsx
```diff
@@ -0,0 +1,17 @@
...
+export function findNotifiableProjectContributors(project: Project): Person[] {
+  const people = project
+    .contributors!.filter((contrib) => contrib.role !== "champion")
+    .map((contrib) => {
+      switch (contrib.role) {
+        case "reviewer":
+          return { ...contrib.person, title: "Reviewer" };
+        default:
+          return { ...contrib.person, title: contrib.responsibility };
+      }
```
shiroyasha, line 13: Let's also cover the case when a non-champion is doing the check-in. It happens in cases when the champion is out-of-office and the team is taking over to cover for them.

lib/operately/notifications/subscription_list.ex
```diff
@@ -9,9 +9,13 @@ defmodule Operately.Notifications.SubscriptionList do
...
 
+  def changeset(attrs) do
+    changeset(%__MODULE__{}, attrs)
+  end
+
   def changeset(subscription_list, attrs) do
     subscription_list
     |> cast(attrs, [:parent_id, :parent_type, :send_to_everyone])
-    |> validate_required([:parent_id, :parent_type])
+    |> validate_required([])
```
shiroyasha, line 19: Why is this not mandatory?

lib/operately_web/api/mutations/post_project_check_in.ex
```diff
@@ -10,6 +10,8 @@ defmodule OperatelyWeb.Api.Mutations.PostProjectCheckIn do
     field :project_id, :string
     field :status, :string
     field :description, :string
+    field :send_notifications_to_everyone, :boolean
+    field :subscribers_ids, list_of(:string)
```
shiroyasha, line 14: `subscriber_ids`?

Rockyy174: Sure, I will change it

assets/js/features/Subscriptions/utils.tsx
Rockyy174: I didn't know there was this possibility. I will make sure to handle it.

assets/js/pages/ProjectCheckInNewPage/page.tsx
```diff
@@ -4,30 +4,39 @@ import * as Pages from "@/components/Pages";
...
   return (
     <Pages.Page title={["Check-In", project.name!]}>
       <Paper.Root>
         <Navigation project={project} />
 
         <Paper.Body>
-          <Form form={form} />
+          <SubscriptionsProvider people={people}>
+            <Form />
+          </SubscriptionsProvider>
```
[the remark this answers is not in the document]
Rockyy174: Yeah, my first instinct was also to put the provider inside the form, but it's not so simple. The `useForm` hook needs information from the provider to submit the data to the backend. So, `useForm` must be inside the provider.

There are a few solutions I can think of, though:

1) Turn `SubscriptionsContext` into a hook. This would solve the problem, but using the Subscription Widget would require a lot more work than it's required now. Currently, we can add the widget to a new form with very few lines of code.

2) I could refactor the whole create-check-in form and update it to use our new form. Then, I could turn the Subscription Widget into a form field. 

3) Remove the `submit` function from the hook and add it to the `SubmitActions` component. The function will have all the data it needs to submit the form and we can put the provider wherever we want, since the provider and the hook will be decoupled completely.


If we must change the current solution, 3) seems to be by far the most straightforward. But let me know what you think.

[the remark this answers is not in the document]
shiroyasha: Thanks for the detailed explanation. I'll have to look into it to to give the best advice.

markoa:
> I'm displaying only the project contributors. Is that correct? Or should I display everyone in the Space or Company instead?

Yes, it is supposed to be the project contributors. By default the notification goes to all contributors and you'd use that option to _narrow down_ the list of people to notify.

Which makes me think..

As a user I'd expect that I can @-mention anyone from the wider circle of people who can see the project but are not contributors, be it from the space or the company. And so if I do that they'll get a notification and that they'll keep receiving notifications about any follow-up comments. Which kind of requires that the "Who to notify widget" becomes aware of them in follow-up comments as well.

markoa:
+1 for having the champion and reviewer on the list but unable to un-check ;)

assets/js/pages/ProjectCheckInNewPage/page.tsx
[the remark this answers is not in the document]
shiroyasha: On the standup we decided that we will use (1).

markoa:
> Currently, if there is only one person to be notified, the text will be: "All 1 people who are contributors of...".

sounds like you need something like
```
(people_list.count > 1) ? "All #{people_list.count} people" : "The 1 person"
```
the second part of the sentence could be simplified to number-neutral `"contributing to #{project.name}"`.

Rockyy174:
Updates:

- I've replaced `SubscriptionsContext` by the `useSubscriptions` hook. Now the subscription data is also handled by the form.

- If there are many people to be notified, the label is "All N people contributing to ${projectName)". 
If it's 1 person, the label is "The 1 person contributing to ${projectName)"

There are still 2 important details missing: 

- I will update the `SubscribersSelector` component so that it can keep a list of people who are always selected. I will use it to keep the Reviewer and Champion always selected.

- I will update the `ProjectCheckIn` operation so that it adds mentioned people to the list. The subscription-list schema already supports it, so I just have to update the operation.

Do you think we should make these changes in this PR or separate ones?

shiroyasha:
Aswm.

> Do you think we should make these changes in this PR or separate ones?

Yes. Let's keep this one simpler.

shiroyasha:
If you are ok with it, I think that we could even merge this PR as it is. 
(you will have to mark it as ready and not a draft)

PR #1022, opened 2024-08-29
chore: Instead of creating a new browser session, clear the cookies to logout

shiroyasha:
My theory is that this tweak will eliminate, or drastically reduce the flakiness related to logins. Why is that?

After running a heavy log-in/log-out loop on my local computer (image down), I've noticed that not every chrome webdriver instance is killed off (I was running htop on the side). This accumulated memory and my machine started to swap which slowed down the execution and ultimately reproduced the same behaviour as we see in flaky tests.

In theory, this should also increase the performance of the test suite and reduce the total time for execution. This is left to be seen in practice.

Heavy log-in/log-out loop on my local computer:

Screenshot 2024-08-29 at 20 20 49

### DIFF
--- a/test/support/features/ui.ex
+++ b/test/support/features/ui.ex
@@ -39,10 +39,23 @@ defmodule Operately.Support.Features.UI do
   def new_session(state) do
     execute(state, fn _ ->
       metadata = Phoenix.Ecto.SQL.Sandbox.metadata_for(Operately.Repo, self())
-      {:ok, session} = Wallaby.start_session([metadata: metadata])
-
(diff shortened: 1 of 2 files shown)


test/support/features/ui.ex
```diff
@@ -39,33 +39,35 @@ defmodule Operately.Support.Features.UI do
...
-    |> end_session()
-    |> new_session()
+    |> logout()
     |> visit(path)
     |> assert_has(testid: "company-home")
   end
 
-  def end_session(state) do
-    if state[:session] do
-      execute(state, fn session ->
```
markoa, line 61: this was surely being executed and still leaving zombies behind?

PR #1040, opened 2024-09-03
feat: Reviewer and Champion always selected in notifications widget

Rockyy174:
In https://github.com/operately/operately/pull/1019, it was mentioned that some people should be always notified and that this problem could be solved in different ways.

In this PR, I solved it by updating the notifications widget so that some people can be never unselected. This is how it looks:

https://github.com/user-attachments/assets/5000c627-8552-4646-b45e-9c757945bfbd

I noticed that it might not be so clear to the user why they can't unselect some people. Do you think it is necessary to add a custom message explaining why the champion and reviewer can't be unselected? And are there any other necessary adjustments?

@shiroyasha @markoa

### DIFF
--- a/assets/js/features/projectCheckIns/Form/useForm.tsx
+++ b/assets/js/features/projectCheckIns/Form/useForm.tsx
@@ -84,7 +84,7 @@ export function useForm({ mode, project, checkIn, author, notifiablePeople = [] 
         status,
         description: JSON.stringify(editor.editor.getJSON()),
         sendNotificationsToEveryone: subscriptionsState.subscriptionType == Options.ALL,
-        subscriberIds: subscriptionsState.selectedPeople.map((person) => person.id!),
+        subscriberIds: subscriptionsState.currentSubscribersList,
       });
 
       navigate(Paths.projectCheckInPath(res.checkIn.id));
--- a/assets/js/features/projectCheckIns/Form/useForm.tsx
+++ b/assets/js/features/projectCheckIns/Form/useForm.tsx
@@ -47,9 +47,11 @@ export interface FormState {
   cancelPath: string;
 }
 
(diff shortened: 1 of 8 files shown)


shiroyasha:
Functionality wise 👌 But yes, we will need a better way to represent why those people can't be deselected.

markoa:
I propose not rendering the checkbox next to the champion and reviewer at all, and next to their "title", adding " - will always be notified".

markoa:
Mainly the idea is that these items should look disabled in the traditional UI sense. Just try a few variations with our predefined colors and go forward with what you think looks best. We can easily tweak later if needed.  On 3. 9. 2024., at 19:10, Adriano Lazzaretti ***@***.***> wrote:﻿

I propose not rendering the checkbox next to the champion and reviewer at all

Makes sense. I will use this approach then.

And use text-content-dimmed to visually communicate the kind of "disabled" state.

Should all the text be in text-content-dimmed (including the person's name)?

Rockyy174:
Here is how it looks now:

https://github.com/user-attachments/assets/f0e42263-b62c-4aee-85a0-0ee5c78e2074

markoa:
LGTM!

assets/js/features/projectCheckIns/Form/useForm.tsx
```diff
@@ -158,3 +160,7 @@ function validate(status: string | null, description: string): Error[] {
 
   return errors;
 }
+
+function getReviewerAndChampion(people: People.Person[]) {
+  return people.filter((p) => p.title === "Reviewer" || p.title === "Champion");
```
shiroyasha, line 165: This confuses me a bit, isn't `person.title` the title of the person in the company?

Rockyy174: Actually, it's displaying the person's role in the project 🤔 

Should I change it to display the title instead?

PR #1047, opened 2024-09-03
chore: Add tooltips to the design system

shiroyasha:
A touchup on the existing tooltip component + adding it to the design system.
Based on Radix primitives: https://www.radix-ui.com/primitives/docs/components/tooltip.

Screenshot 2024-09-04 at 01 51 46


shiroyasha:
It didn't look good in dark mode because it blended in the background. I've added a subtle stroke around to mitigate the problem:

Screenshot 2024-09-04 at 02 13 02

PR #1078, opened 2024-09-06
feat: Add subscriptions "Add/remove people..." button

Rockyy174:
Now it's possible to edit a subscriptions list.

https://github.com/user-attachments/assets/109f991c-8bbc-42b5-85b7-9d680308bb58


assets/js/features/Subscriptions/selector/SubscribersSelectorForm.tsx
```diff
@@ -0,0 +1,62 @@
...
+        <div className="flex items-center gap-3">
+          <ActionLink onClick={handleSelectEveryone}>Select everyone</ActionLink>
+          <span className="text-content-dimmed">&middot;</span>
+          <ActionLink onClick={handleSelectNoone}>Select no one</ActionLink>
+        </div>
+
+        <Forms.MultiPeopleSelectField field="people" />
+      </div>
+
+      <div className="flex justify-center gap-2">
```
shiroyasha, line 52: If we add a `cancelText?: string` to Forms.Submit, can we use it here?

assets/js/features/Subscriptions/current-subscriptions/ExistingSubscriptionsList.tsx
```diff
@@ -0,0 +1,48 @@
...
+  switch (subscriptions.length) {
+    case 0:
+      prefix = "No one";
+      break;
+    case 1:
+      prefix = "1 person";
+      break;
+    default:
+      prefix = `${subscriptions!.length} people`;
+  }
```
shiroyasha, line 45: This can be a bit more nice with `ts-pattern`. Example:

``` tsx
let prefix = match(subscriptions.length)
  .with(0, () => "No one")
  .with(1, () => "1 person")
  .otherwise(() => `${subscriptions.length} people`);
```

Rockyy174: Cool. This syntax is new to me. Next time I touch this file, I will make sure to update it.

assets/js/features/Subscriptions/selector/SubscribersSelectorForm.tsx
Rockyy174: I'm not sure if I understood. 
Do you mean adding a prop `cancelText?` so that we can have different text in the cancel button instead of the "Never mind" text?

shiroyasha: No, I meant that we have a compoenent called `<Forms.Submit saveText="Submit This" />` that rendered a submit area for the form. If we extend it with the cancelText prop `<Forms.Submit saveText="Submit This" cancelText="Nevermind" />` it would be able to rended a "default" version of the submit area all around the app.

PR #1097, opened 2024-09-09
feat: Creating check-in sends notifications using subscriptions list

Rockyy174:
I've updated `Operately.Activities.Notifications.ProjectCheckInSubmitted` so that it uses the check-in's subscriptions_list to send notifications. The new behavior is:

- If the notifications_list has `send_to_everyone = false`, it sends a notification to everyone who has a subscription (not canceled).

- If the notifications_list has `send_to_everyone = true`, it sends a notification to every contributor, unless the contributor's subscription was canceled.

I haven't handled permissions yet.

I've also added the module `Operately.Projects.Notifications`. My plan is to add to this module all the functions that will be used to get a resource's subscribers. So far, I've added a function to get a check-in's subscribers.

@shiroyasha, What do you think about the new notifications behavior? And how about the `Operately.Projects.Notifications` module?

### DIFF
--- a/lib/operately/activities/notifications/project_check_in_submitted.ex
+++ b/lib/operately/activities/notifications/project_check_in_submitted.ex
@@ -1,23 +1,15 @@
 defmodule Operately.Activities.Notifications.ProjectCheckInSubmitted do
-  alias Operately.Projects
+  alias Operately.{Notifications, Projects}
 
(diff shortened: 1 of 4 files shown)


shiroyasha:
Looks good 👍 I like the new module.

PR #1099, opened 2024-09-10
feat: CommentAdding operation creates subscriptions

Rockyy174:
I've updated `Operately.Operations.CommentAdding` so that, if someone is mentioned in a new comment, a notifications subscription is created for this person.


Rockyy174:
Unfortunately, I don't know. Actually, I didn't know it was only my pull requests that were failing. 
I noticed that docker build works after we merge the pull request, though 🤔

PR #1100, opened 2024-09-10
chore: Migration for creating subscriptions list for existing check-ins

Rockyy174:
I've added a data migration which creates a subscriptions list for existing project check-ins. A subscription is created for every contributor of the project.

If a check-in already has a subscriptions list or a contributor already has subscription, they are skipped.


shiroyasha:
We have some conflicts after the merge.

Rockyy174:
I've just fixed the conflicts.

PR #1118, opened 2024-09-12
chore: Migrate Goal Updates to its own table

Rockyy174:
I've added migration files to create a new table for Goal Updates. The migrations should also handle the copying of existing data.

I've also updated our existing queries and mutations to use the new table.


lib/operately/goals.ex
```diff
@@ -97,4 +97,48 @@ defmodule Operately.Goals do
...
+      join: g in assoc(b, :group),
+      join: m in assoc(g, :memberships),
+      join: p in assoc(m, :person),
+      where: m.person_id == ^person_id and is_nil(p.suspended_at),
+      where: b.access_level >= ^Binding.view_access(),
+      where: u.id == ^id,
+      preload: [:author, :acknowledged_by, reactions: [:person], goal: [:targets]],
+      group_by: [u.id, u.goal_id, u.author_id, u.message, u.acknowledged_at, u.acknowledged_by_id, u.targets, u.inserted_at, u.updated_at],
+      select: {u, max(b.access_level)}
+    )
```
Rockyy174, line 137: This is a temporary query which should be removed as soon as possible (maybe even before merging this PR).

The thing is that the functions in both `Operately.Repo.Getter` and `Operately.Access.Fetch` wouldn't work with the new `Operately.Goals.Update`. Ecto kept complaining that `group_by` had to include all the fields.

Probably I've added something different to the new schema, but I can't tell what it is. What I know is that, unless I include all the fields in the aggregation, Ecto will throw an error.

Do you know what could be causing this problem, @shiroyasha?

shiroyasha: > If a table is structured in such a way that the GROUP BY column is also the primary key or a unique identifier (like r.id), PostgreSQL might "assume" that all other columns from the table are functionally dependent on this GROUP BY column.
> 
> In these cases, PostgreSQL allows you to select other fields from the table without explicitly including them in the GROUP BY clause. This works because PostgreSQL knows there can only be one unique row per r.id.
> 
> However, if the table does not have this kind of uniqueness or functional dependency, PostgreSQL will throw an error, requiring that all non-aggregated fields be included in the GROUP BY clause.

In the new table, is the ID field still marked as a primary key? This might be the source of the problem. If not, we can review it together in the morning.

shiroyasha: To confirm, yes, this will be the most likely cause:

Screenshot 2024-09-12 at 23 39 14

Rockyy174: Yes, this was the problem. Thank you! 🙇‍♂️

PR #1125, opened 2024-09-14
feat: Add subscriptions list to goal update

Rockyy174:
I've updated the create-goal-update form and `Operately.Operations.GoalCheckIn` so that creating a goal update also creates a subscriptions list.

I noticed that creating a subscriptions list for goal updates and project check-ins is very similar. And I'm pretty sure it's going to be as similar for other resources in the future. So, I did some small refactoring to use the same functions in both operations. 

Instead of: 
- `Operately.Operations.ProjectCheckIn.Subscription`
- `Operately.Operations.ProjectCheckIn.SubscriptionList`

Now we have:
- `Operately.Operations.Notifications.Subscription` 
- `Operately.Operations.Notifications.SubscriptionList`

Also, creating goal updates in tests was a bit cumbersome, so I've added `goal_update_fixture`.

### DIFF
--- a/assets/js/features/Subscriptions/utils.tsx
+++ b/assets/js/features/Subscriptions/utils.tsx
@@ -1,5 +1,6 @@
 import { Person } from "@/models/people";
 import { Project } from "@/models/projects";
+import { Goal } from "@/models/goals";
 import { Subscription } from "@/models/notifications";
 import { NotifiablePerson } from "@/features/Subscriptions";
 import { compareIds, includesId } from "@/routes/paths";
@@ -8,6 +9,10 @@ export function getSelectedPeopleFromSubscriptions(people: NotifiablePerson[], s
   const ids = subscriptions.map((s) => s.person!.id!);
 
   return people.filter((p) => includesId(ids, p.id));
+}
+
+export function getReviewerAndChampion(people: NotifiablePerson[]) {
+  return people.filter((p) => p.title === "Reviewer" || p.title === "Champion");
 }
 
 export function findNotifiableProjectContributors(project: Project, me?: Person): NotifiablePerson[] {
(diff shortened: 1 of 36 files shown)


assets/js/features/Subscriptions/utils.tsx
```diff
@@ -37,3 +42,27 @@ export function findNotifiableProjectContributors(project: Project, me?: Person)
...
+        fullName: member.fullName!,
+        avatarUrl: member.avatarUrl!,
+        title: member.title!,
+      };
+
+      if (compareIds(member.id, goal.reviewer?.id)) {
+        person.title = "Reviewer";
+      }
+      if (compareIds(member.id, goal.champion?.id)) {
+        person.title = "Champion";
```
shiroyasha, line 61: I've found myself again lost with why do we change the title. Can we change this to `role` on the NotifiablePerson struct?

lib/operately_web/api/serializers/goal.ex
```diff
@@ -37,4 +37,15 @@ defimpl OperatelyWeb.Api.Serializable, for: Operately.Goals.Goal do
...
+  defp serialize_space(space) do
+    cond do
+      not Ecto.assoc_loaded?(space) ->
+        OperatelyWeb.Api.Serializer.serialize(space)
+      Ecto.assoc_loaded?(space.members) and Ecto.assoc_loaded?(space.company) ->
+        OperatelyWeb.Api.Serializer.serialize(space, level: :full)
+      true ->
+        OperatelyWeb.Api.Serializer.serialize(space)
+    end
+  end
```
shiroyasha, line 50: Is this a possible simplification:

``` elixir
defp serialize_space(space) do
  if Ecto.assoc_loaded?(space) && Ecto.assoc_loaded?(space.members) do
     Serializer.serialize(space, level: :full)
  else
     Serializer.serialize(space)
  end
end
```

shiroyasha:
Awesome. I like that we have simplified the subscription modules to a single unified one 👏

lib/operately_web/api/serializers/goal.ex
Rockyy174: Yes, I think this may be cleaner than the current 'cond' solution. 

One thing to keep in mind, though, is that the serializer also needs the company. 

If, for some reason, someone ended up loading members, but not the company, the serializer would break. So, the `if` statement would be quite long because it has to check if company is also loaded:

```
if Ecto.assoc_loaded?(space) && Ecto.assoc_loaded?(space.members) && and Ecto.assoc_loaded?(space.company) do
```

shiroyasha: Writing it down to think about this. It would be great to have a general solution to this.

Issue #836, opened 2024-08-05
API: Dates and Datetimes are not automatically converted to Date types


shiroyasha:
Manually converting date types in the API response is PITA and leaves room for subtle bugs like. The automation around this should be straightforward and would help us eliminate unnecessary work.

PR #1181, opened 2024-09-23
Created the DateTimeHelpers

prathameshjadhav73:
Addressing issue #836,

How implemented this automated solution: 

1. Created the `DateTimeHelpers` file.
2. Updated only one serializer file `goal.ex` to use the `DateTimeHelpers.serialize_datetime` function for date and datetime fields, for showcase.
3. Also works for serializers that might contain complex nested structures like `list.ex`.

### DIFF
--- a/lib/operately_web/api/serializers/date_time_helpers.ex
+++ b/lib/operately_web/api/serializers/date_time_helpers.ex
@@ -1,0 +1,7 @@
+defmodule OperatelyWeb.Api.Serializers.DateTimeHelpers do
+  def serialize_datetime(nil), do: nil
+  def serialize_datetime(%DateTime{} = datetime), do: DateTime.to_iso8601(datetime)
+  def serialize_datetime(%NaiveDateTime{} = naive_datetime), do: NaiveDateTime.to_iso8601(naive_datetime)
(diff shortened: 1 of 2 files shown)


shiroyasha:
Hey @prathameshjadhav73!

Thanks for jumping in with this! Always great to see first-time contributions.

That said, we’re currently leaning on a protocol-based approach for this kind of serialization, which gives us more flexibility across serializers and works well with complex nested structures. Here’s a link to the protocol-based solution we’re using: https://github.com/operately/operately/blob/main/lib/operately_web/api/serializers/any.ex#L4.

I’ll take the blame for the original issue(https://github.com/operately/operately/issues/836) not being well explained. The real problem is actually in the generated API on the JavaScript side. For instance, when we call `Api.getGoal({ id })`, the backend sends the correct data, but the date fields come through as strings. Ideally, the generated API stubs should convert these strings into proper `Date` objects, so we don’t have to manually parse the in the React application. That’s the problem we’re trying to solve.

Appreciate the effort though! If you’re interested, I’d love to see a PR tackling the JS API generation issue instead, I can also provide some tips where to start looking for a solution in out codebase.

shiroyasha:
I'll close this PR @prathameshjadhav73 as it is stale. Let me know if you want help with tackling #836.

PR #1183, opened 2024-09-24
feat: Make Project Retrospective editable

### LINKED ISSUE #685: Make project retrospectives editable
It's not cool when you can't change what you just posted.

Visual direction: let's reuse the message layout. Wider box with (...) menu to access the Edit action, boost widget, and comments below.

Rockyy174:
A Project Retrospective is now editable. 
Solves https://github.com/operately/operately/issues/685.


shiroyasha:
I have one thing that bit me while working on project contributor editing: If the user opens the page, makes no changes, and clicks save, we should avoid creating an empty activity feed item. I would like to avoid introducing any more update actions in the API that behave like that.

One way to solve this is on the frontend: When clicking the save button we can check if there are changes, and if not, we can just redirect back.

Alternatively, we could check in on the backend, by comparing pervious/new data, and simply returning 200 without making any modifications.

Rockyy174:
Makes sense. With the project contributors page, did you take the backend or frontend approach?

shiroyasha:
I went with the frontend approach, but to be honest, I'm not sure that I've made the right decision. If we can, I would go with the backend one.

Rockyy174:
What do you think about this approach, @shiroyasha?

I wasn't sure if I should handle it at the mutation or operation level. I went with the operation level because we might use the operation somewhere else in the application.

shiroyasha:
Looks good 👍

PR #1197, opened 2024-09-26
fix: Subscriptions updated to work with our new forms

Rockyy174:
It seems that the latest update in forms broke the subscriptions widget. This PR fixes it.


shiroyasha:
Thanks for fixing it! Side-note: We will need some integration tests for this.

Rockyy174:
Yeah, I will make sure to add integration tests before closing this project.

Issue #1206, opened 2024-09-27
There is no way to abandon/cancel a project


shiroyasha:
Writing down a feedback I got from the Semaphore team:

In some cases a project might be started but due to shifts in priority it has to be canceled.
Currently Operately allows closing a project, but there is no way to mark it as canceled/abandoned/not-even-started. To make it worse, currently we use the status `Completed` coupled with a green checkmark.

![Screenshot 2024-09-27 at 16 32 46](https://github.com/user-attachments/assets/834aef23-cc57-4fd6-8692-4d61c6e2203b)

A quick fix is to use a neutral term `Closed`, but it is just a patch because ideally we would display what really happened with the project: It was canceled.

cc-ing @markoa as we could incorporate this change into the CCP that introduced the new Goal Tree.

PR #1259, opened 2024-10-03
Project cancellation option added.

AnjaliKamble:
Referring to #1206, I have added new files which help us with cancellation of groups, cancellation of goals and cancellation of projects. Upon further refinement of the code, we can expect it to fully function without any bugs and errors.


shiroyasha:
Thank you for the PR @AnjaliKamble but we are only looking for project cancelation, not group or goal cancelation. Please remove the other two variants and we can continue from there.

PR #1279, opened 2024-10-07
feat: Notifications are cleared when discussion comments are seen by user

Rockyy174:
When a user sees a comment in a discussion, if there is an unread notification for this comment, the notification is marked as read.

https://github.com/user-attachments/assets/0d5d2947-e4bd-4cda-beb3-df7399fa500c

### DIFF
--- a/assets/js/features/CommentSection/CommentSection.tsx
+++ b/assets/js/features/CommentSection/CommentSection.tsx
@@ -8,6 +8,7 @@ import Avatar from "@/components/Avatar";
 import FormattedTime from "@/components/FormattedTime";
 import RichContent from "@/components/RichContent";
 import { PrimaryButton, SecondaryButton } from "@/components/Buttons";
+import { useClearNotificationOnIntersection } from "@/features/notifications";
 
 import { FormState } from "./form";
 import { useBoolState } from "@/utils/useBoolState";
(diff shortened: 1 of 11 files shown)


assets/js/features/CommentSection/CommentSection.tsx
```diff
@@ -165,15 +166,48 @@ function MilestoneReopened({ comment }) {
...
+    if (commentRef.current && comment.notification && !comment.notification.read) {
+      // setTimeout prevents the callback function from being fired when IntersectionObserver
+      // is instantiated, which is its default behavior.
+      setTimeout(() => observer.observe(commentRef.current!), 500);
+    }
+
+    return () => {
+      observer.disconnect();
+    };
+  }, []);
```
shiroyasha, line 204: The ref should be in the deps list, right? (I'm 80% sure)

assets/js/features/CommentSection/CommentSection.tsx
```diff
@@ -165,15 +166,48 @@ function MilestoneReopened({ comment }) {
...
+  const [markNotificationAsRead] = useMarkNotificationAsRead();
+
   const entity = { id: comment.id, type: "comment", parentType: commentParentType };
   const addReactionForm = useReactionsForm(entity, comment.reactions);
+
   const testId = "comment-" + comment.id;
   const content = JSON.parse(comment.content)["message"];
+  const commentRef = React.useRef<HTMLDivElement>(null);
+
+  React.useEffect(() => {
```
shiroyasha, line 178: Exrtact this to dedicated hook. Its too complex to live inside of this function.

Rockyy174: Makes sense!

assets/js/features/CommentSection/CommentSection.tsx
Rockyy174: I'm not so sure. But if I'm not mistaken, a ref never triggers a re-render, so it shouldn't be added to the array.

I found the following article on this topic:
https://www.epicreact.dev/why-you-shouldnt-put-refs-in-a-dependency-array

PR #1325, opened 2024-10-13
issue with notification count

Virendrapro:
To solve the issue of inconsistent unread notification counts in the Operately tests, the following steps were implemented:

Notification Reset: Added a call to Operately.Notifications.clear_all() in the setup block to ensure a clean state for notifications before each test.

Session Management: Incorporated UI.logout() at the end of each feature test to clear user sessions, preventing carry-over of states between tests.


shiroyasha:
Hey @Virendrapro! Thanks for making your first PR.

 I didn't know about this problem:

> To solve the issue of inconsistent unread notification counts in the Operately tests

Can you show an example where this happens?

PR #1328, opened 2024-10-14
fix: Make discussions mobile responsive

markoa:
before:
Screenshot 2024-10-14 at 10 45 41

after:
Screenshot 2024-10-14 at 10 45 52

### DIFF
--- a/assets/js/components/PaperContainer/Navigation.tsx
+++ b/assets/js/components/PaperContainer/Navigation.tsx
@@ -5,7 +5,7 @@ import { Link } from "@/components/Link";
 
 export function Navigation({ children }) {
   return (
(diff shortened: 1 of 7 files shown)


shiroyasha:
@markoa as this touches every page in operately, please double check if every major page still looks good.

shiroyasha:
Second thing: IMHO we don't need margins around the page on mobile. Snap it to the border of the screen.

markoa:
> please double check if every major page still looks good.

Thanks for the reminder. I realized I haven't checked Goals on desktop and there was a glitch that I fixed by removing some unnecessary changes in PaperContainer.

> we don't need margins around the page on mobile. Snap it to the border of the screen.

Many changes here reduce margins on mobile, I'd postpone snapping content to another PR.

markoa:
Linter caught a kind of a type error that I fixed, the build is green now.

PR #1341, opened 2024-10-14
fix: Normalize padding in PaperContainer body

### LINKED ISSUE #1336: Extra page paddings on space page
Extra top padding appeared on the space page. Most likely after this change: https://github.com/operately/operately/pull/1328.

Screenshot 2024-10-14 at 14 52 24

markoa:
Fixes #1336.

### DIFF
--- a/assets/js/components/PaperContainer/index.tsx
+++ b/assets/js/components/PaperContainer/index.tsx
@@ -63,10 +63,28 @@ export function Root({ size, children, className, fluid = false }: RootProps): J
   );
 }
 
-export function Body({ children, minHeight = "none", className = "", backgroundColor = "bg-surface" }) {
+const bodyPaddings = {
+  small: "px-10 py-8",
+  medium: "px-12 py-10",
+  large: "px-12 py-10",
+  xlarge: "px-12 py-10",
+  xxlarge: "px-16 py-12",
+};
+
+export function Body({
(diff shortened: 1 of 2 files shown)


assets/js/components/PaperContainer/index.tsx
```diff
@@ -63,10 +63,28 @@ export function Root({ size, children, className, fluid = false }: RootProps): J
...
+};
+
+export function Body({
+  children,
+  minHeight = "none",
+  className = "",
+  noPadding = false,
+  backgroundColor = "bg-surface",
+}) {
+  const { size } = React.useContext(Context);
```
markoa, line 81: I tried to replace this with a tailwind one-liner but couldn't make it work

shiroyasha: What was the limiting factor?

shiroyasha:
Awesome! Thanks for the quick fix. Shipping 🚢

assets/js/components/PaperContainer/index.tsx
markoa: if I'd translate the paddings hash into `sm:xxx lg:zzz` etc then it would work on 1-2 screen sizes but not on the rest, and it have side effects like the top of spaces nav bar would get out of position etc.

Issue #751, opened 2024-07-11
Show time for posted messages


shiroyasha:
example: Opening the following gives me no clue if this was posted in the morning, or just a couple of minutes ago.

![Screenshot 2024-07-11 at 13 16 00](https://github.com/operately/operately/assets/1779493/01992ab1-3ab4-4fe8-a11b-791efbdf8cb2)

PR #1345, opened 2024-10-15
feat: Improve how discussion posted time is displayed

Rockyy174:
Fixes https://github.com/operately/operately/issues/751.

First, I thought about displaying both the date and time when the discussion was posted, but it didn't look good. 

So, the approach I took was to render the `FormattedTime` component with `format="relative"` if the discussion was posted less than 24 hours ago, but render it with `format="short-date"` if the discussion was posted more than 24 hours. 

Here is how it looks:

https://github.com/user-attachments/assets/cf549698-59f6-4e19-adf9-63c4edb08639

### DIFF
--- a/assets/js/pages/DiscussionPage/page.tsx
+++ b/assets/js/pages/DiscussionPage/page.tsx
@@ -22,6 +22,7 @@ import { Paths, compareIds } from "@/routes/paths";
 import { CurrentSubscriptions } from "@/features/Subscriptions";
 import { useClearNotificationsOnLoad } from "@/features/notifications";
 import { assertPresent } from "@/utils/assertions";
+import { hoursBetween, parse } from "@/utils/time";
 
 export function Page() {
   const me = useMe()!;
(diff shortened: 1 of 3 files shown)


markoa:
Is this rendered "live", I mean if I open the page and it says "Posted just now", leave my desk and come back in 15 minutes, will it way "Posted 15 minutes ago"?

Rockyy174:
I don't think so. The "posted time" only updates when the data is refreshed. However, we could find a way to refresh the time, such as on a time interval or when the page is focused, etc.

The only problem I see is that the "posted time" is loaded with the whole discussion + a lot of other resources, so all the data would be refreshed, not only the "posted time", which doesn't seem very efficient.

Unless an API endpoint is created only for loading the "posted time" 🤔

shiroyasha:
Let's merge as it is. I have an idea how to make relative time auto-update with a relatively simple approach that would work for the whole app.

I'll share a proof of concept, and if it is viable you can take over.

shiroyasha:
Here is the Proof-Of-Concept: https://github.com/operately/operately/pull/1346.

PR #1346, opened 2024-10-15
feat: Proof-of-Concept: Auto-refresh relative time

shiroyasha:
Warning: Not fully tested, just on this example.

- [ ] some tweaks are needed for the refresh interval (not too fast, not too slow)
- [ ] a phase transition for when discussion time changes from relative time -> short-date

The current behaviour with refresh time 10seconds.

https://github.com/user-attachments/assets/573846e6-28bb-4fc3-a397-3ea75fe9dfc8


markoa:
Cool! IMO a 1-minute refresh resolution should be good enough. In the context of business and modern apps displaying potentially wrong time is not a small issue so as long we don't do that we should be good.

PR #1348, opened 2024-10-15
feat: RelativeTime refreshes time on interval

Rockyy174:
Implementation of https://github.com/operately/operately/pull/1346.

Now, whenever we use `<FormattedTime time={time} format="relative" />`, the component will refresh on a specific interval. The refresh interval is shorter or longer depending on how long ago `time` is:
- if time < 1 minute, the interval is 15 seconds
- else if time < 1 hour, the interval is 60 seconds
- else if time < 1 day, the interval is 60 minutes
- else, the interval is a day

I've also added the format option `relative-time-or-date` to the `FormatTime` component, which displays the same result as ` format="relative"` if `time` is less than 24 hours ago, otherwise it displays the same result as ` format="relative-time-or-date"`.

### DIFF
--- a/assets/js/components/FormattedTime/useRenderInterval.tsx
+++ b/assets/js/components/FormattedTime/useRenderInterval.tsx
@@ -1,0 +1,47 @@
+import { useState, useEffect } from "react";
+
+export function useRenderInterval(time) {
+  const [lastRender, setLastRender] = useState(0);
+
+  useEffect(() => {
+    const diff = +new Date() - +time;
+
+    const seconds = Math.floor(diff / 1000);
+    const minutes = Math.floor(seconds / 60);
+    const hours = Math.floor(minutes / 60);
+
+    if (seconds < 60) {
+      const interval = setInterval(() => {
+        setLastRender(Date.now());
+      }, 1000 * 15);
+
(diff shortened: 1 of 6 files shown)


assets/js/components/FormattedTime/useRenderInterval.tsx
```diff
@@ -0,0 +1,47 @@
+import { useState, useEffect } from "react";
+
+export function useRenderInterval(time) {
+  const [lastRender, setLastRender] = useState(0);
+
```
Rockyy174, line 5: After analyzing the code, I realized that one possible optimization in the hook could be to make sure that the interval changes as the `diff` grows. 

For example, if we open a several-hour-old discussion, the interval will be 1 hour, which is good.

However, if we create a discussion and leave the page open, the interval will always be 15 seconds until the page is reloaded.

If we were making requests, I'd say that this optimization would be definitely necessary. But since we are not doing any heavy stuff, I'm not so sure. What do you think @shiroyasha?

shiroyasha: The current implementation is good enough. Additional micro optimization could be made if we find it necessary, later.

PR #1358, opened 2024-10-17
feat: Add Goals V2 page

Rockyy174:
I've added the Goals V2 page. 

tmp

So far, the page only shows the existing Goals and Projects and the only functionality that is supported is collapse/expand a Goal's children.

The only way to access this page is to manually go to `/company-id/goals-v2`. The idea here is to add to the page all features it needs, test it well, then swap it with the old Goals page.


shiroyasha:
We now have two different concepts called goal-v2 😅

PR #1406, opened 2024-10-22
chore: Signals

shiroyasha:


Rockyy174:
I signed off the commit in the original branch (https://github.com/operately/operately/pull/1409), but it seems it somehow got lost when I merged the commit 🤔

PR #1418, opened 2024-10-23
chore: Add UI tests for Goals V2 page

Rockyy174:
I've added UI tests for the new Goals page.

### DIFF
--- a/test/features/goal_tree_test.exs
+++ b/test/features/goal_tree_test.exs
@@ -4,7 +4,8 @@ defmodule Operately.Features.GoalTreeTest do
   alias Operately.Support.Features.GoalTreeSteps, as: Steps
 
   setup ctx do
-    {:ok, Steps.setup(ctx)}
+    ctx
+    |> Steps.given_goals_and_projects_exist()
   end
 
   feature "show/hide completed projects", ctx do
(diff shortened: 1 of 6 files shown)


test/features/goal_tree_test.exs
```diff
@@ -15,4 +16,77 @@ defmodule Operately.Features.GoalTreeTest do
...
+    |> Steps.apply_filter(["ownedBy-anyone", "reviewedBy-me"])
+    |> Steps.assert_project_visible(:goal_1)
+    |> Steps.refute_project_visible(:goal_2)
+  end
+
+  #
+  # Helpers
+  #
+
+  defp visit_goals_v2_tmp(ctx) do
```
shiroyasha, line 88: Move this to steps as well. In general let's aim to have most (all?) steps in the steps file.

Rockyy174: Makes sense. Actually, I intend to delete this function as soon as we replace the goals page, so I didn't pay much attention to it. But I will move it to the steps file anyway.

test/features/goal_tree_test.exs
```diff
@@ -15,4 +16,68 @@ defmodule Operately.Features.GoalTreeTest do
...
+    |> Steps.assert_project_visible(:project_alpha)
+    |> Steps.assert_project_visible(:project_beta)
+    |> Steps.assert_project_visible(:goal_2)
+    |> Steps.show_filter_options()
+    |> Steps.apply_filter(["ownedBy-me"])
+    |> Steps.refute_project_visible(:goal_1)
+    |> Steps.show_filter_options()
+    |> Steps.apply_filter(["ownedBy-anyone", "reviewedBy-me"])
+    |> Steps.assert_project_visible(:goal_1)
+    |> Steps.refute_project_visible(:goal_2)
```
shiroyasha, line 81: Feature tests give more value if they are written from the users's perspective, which in practice means using statements that describe the feature from the perspective of the user. Here is the main benefit: When someone else reads the steps, he should have a good understanding of the user interaction and the expection of the user.

In your example, you are focusing on what is happening, but it is not clear why.

```
ctx
|> visit_goals_v2_tmp()
|> Steps.assert_project_visible(:project_alpha)
|> Steps.assert_project_visible(:project_beta)
|> Steps.expand_collapse_node(id: Paths.goal_id(ctx.goal_1))
|> Steps.refute_project_visible(:project_alpha)
|> Steps.refute_project_visible(:project_beta)
|> Steps.expand_collapse_node(id: Paths.goal_id(ctx.goal_1))
|> Steps.assert_project_visible(:project_alpha)
|> Steps.assert_project_visible(:project_beta)
```

Here is an alternative that focuses on the user journey:

```
|> Step.given_the_user_is_a_company_member()
|> Step.when_the_user_visits_the_goalsv2_page()
|> Step.then_all_projects_are_visible_by_default()
|> Step.when_the_user_expands_a_goal()
|> Step.then_subgoals_and_projects_of_that_goal_should_be_visible()
```

There steps are using using [the given/when/then statements](https://martinfowler.com/bliki/GivenWhenThen.html) popular in many testing frameworks, especially in ones that are focusing on [BDD](https://cucumber.io/docs/bdd/). 

Using alternative kinds of steps (i.e. not being dogmatic about given/when/then) is also OK but we should aim to keep in on the high-level, and from the users perspective.

For example:

```
|> Step.given_iam_a_company_member()
|> Step.visit_the_goalsv2_page()
|> Step.assert_all_projects_are_visible_by_default()
|> Step.expands_a_goal()
|> Step.assert_that_subgoals_and_projects_of_that_goal_are_visible()
```

test/support/features/goal_tree_steps.ex
```diff
@@ -38,4 +45,48 @@ defmodule Operately.Support.Features.GoalTreeSteps do
   step :assert_project_visible, ctx, id do
     UI.assert_text(ctx, ctx[id].name)
   end
+
+  step :expand_collapse_node, ctx, id: id do
+    testid = "toggle-node-" <> id |> String.downcase()
```
shiroyasha, line 50: There is a helper: `UI.testid(...)` that can help with this.

shiroyasha (changes requested):

test/features/goal_tree_test.exs
Rockyy174: Got it. I will make sure to update the step names.

shiroyasha: Nice changes, the new ones are much better. The only thing that remains is the `filter goals and projects` feature. It is quite intimidating block of tests. Can you brake it down into smaller features so that it is more manageable?

PR #1425, opened 2024-10-24
feat: Add Discussions to space tools

Rockyy174:
I've started adding the new ToolsSection to the Space page. As this is going to be too much code for a single PR, I decided to build it in the `features/SpaceTools/` directory and add the directory to the ignore list in the `knip.json` file. This way the linter won't complain. 

To see how the component looks, we just need to add it to the Space page:

```
<ToolsSection space={space} />
```

I started building it by the Discussions tool. Here is how it looks:

tmp

I'm marking this PR as a draft because I think we could improve some things before merging it:

1) I'm using `Paper.DimmedSection`, that's why the background is not white as in the wireframe. The reason why I'm using it is because it calculates the negative margins, which is necessary for adding a new section to the page. I didn't manage to find a section component in the codebase that has a white background. Do we have one? If not, I can create one that behaves similarly to `Paper.DimmedSection`, but with another background.

2) Currently, while the discussions are loading, I just hide the component. Do we have a spinner component? If not, I can also create one.

3) What do you think about the `surface-outline` background color in the comments count? Should I try another color?

@shiroyasha @markoa

### DIFF
--- a/assets/js/api/index.tsx
+++ b/assets/js/api/index.tsx
@@ -629,6 +629,7 @@ export interface Discussion {
   space?: Space | null;
   reactions?: Reaction[] | null;
   comments?: Comment[] | null;
+  commentsCount?: number | null;
   subscriptionList?: SubscriptionList | null;
   potentialSubscribers?: Subscriber[] | null;
   notifications?: Notification[] | null;
@@ -1318,6 +1319,8 @@ export interface GetDiscussionResult {
 
 export interface GetDiscussionsInput {
   spaceId?: string | null;
+  includeAuthor?: boolean | null;
+  includeCommentsCount?: boolean | null;
 }
 
 export interface GetDiscussionsResult {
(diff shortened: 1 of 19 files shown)


shiroyasha:
Nice! This feature is my personal favorite, can't wait to see it in production.

Answering your questions:

1. Do we need the negative margins? I think if we leave the section below the title in a centered layout it would look good. No separator or anything fancy.

2. Let's avoid spinners if possible, especially for navigation elements which these boxes are. They are reducing the user experience, and introducing unnecessary friction 🙏 Let's focus on making the API fast and load the data as part of the page loader.

3. The contrast between test and foreground is not the best. I vote for some "active" non-gray color, like blue bg + white text. For example, let's try `bg-blue-500` + `text-white-1`. We can do this in another PR as well, as I assume it will take a couple of iterations to get it right.

Rockyy174:
Here is what it looks like now:

tmp

PR #1432, opened 2024-10-28
feat: Display goals and projects headers in space tools

Rockyy174:
I've added a pie chart component which will be used in the Goals & Projects header in the space tools.

### DIFF
--- a/assets/js/features/SpaceTools/GoalsAndProjects.tsx
+++ b/assets/js/features/SpaceTools/GoalsAndProjects.tsx
@@ -1,11 +1,65 @@
 import React from "react";
 
-import { Container, Title } from "./components";
+import { Goal } from "@/models/goals";
+import { Project } from "@/models/projects";
 
-export function GoalsAndProjects() {
+import { Container, Title } from "./components";
+import { calculateGoalsStatus, calculateProjectsStatus } from "./utils";
+import { PieChart } from "@/components/PieChart";
+
+interface GoalsAndProjectsProps {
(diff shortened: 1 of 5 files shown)


assets/js/features/SpaceTools/GoalsAndProjects.tsx
```diff
@@ -1,11 +1,65 @@
...
+}
+
+function Header(props: GoalsHeader | ProjectsHeader) {
+  const status: ResourceStatus = React.useMemo(() => {
+    switch (props.type) {
+      case "goals":
+        return calculateGoalsStatus(props.goals);
+      case "projects":
+        return calculateProjectsStatus(props.projects);
+    }
```
shiroyasha, line 49: Use `match`. The benefit is that if it doesn't match the type properly it will raise an exception, while the switch is silently passing on an undefined field.

Rockyy174: I remember that I tried to use `match`, but it seems that it has some limitations when you work with union types. When I use `case`, it understands that, if `type == "goals"`, then only `props.goals` can be used, and if `type == "projects"`, then only `props.projects` can be used. But `match` doesn't understand that.

I've just tried it again. This what it looks like:

tmp


Do you think I may be using `match` incorrectly?

shiroyasha: It would probably require type assertions. Merging it as it is.

PR #1438, opened 2024-10-28
feat: Space tools are responsive + hovering on them shows shadow

Rockyy174:
When we hover over a tool, a subtle shadow appears:

https://github.com/user-attachments/assets/5199f1c4-db17-4f0d-81ea-0f29508c225f

Also, space tools are now responsive. 

2 tools:
https://github.com/user-attachments/assets/053f7c87-c6c0-48f0-ac68-ebc63013e7c5

5 tools: 
https://github.com/user-attachments/assets/d78f1155-694a-41f2-abd8-bd1dfbdb4fc9

### DIFF
--- a/assets/js/features/SpaceTools/ToolsSection.tsx
+++ b/assets/js/features/SpaceTools/ToolsSection.tsx
@@ -19,8 +19,8 @@ export function ToolsSection({ space, discussions, goals, projects }: ToolsSecti
   return (
     <div className="mt-6 py-6">
       <div className="flex justify-center items-start flex-wrap gap-8">
-        <GoalsAndProjects space={space} goals={goals} projects={projects} />
-        <Discussions space={space} discussions={discussions} />
+        <GoalsAndProjects space={space} goals={goals} projects={projects} toolsCount={2} />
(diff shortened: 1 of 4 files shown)


assets/js/features/SpaceTools/ToolsSection.tsx
```diff
@@ -19,8 +19,8 @@ export function ToolsSection({ space, discussions, goals, projects }: ToolsSecti
   return (
     <div className="mt-6 py-6">
       <div className="flex justify-center items-start flex-wrap gap-8">
-        <GoalsAndProjects space={space} goals={goals} projects={projects} />
-        <Discussions space={space} discussions={discussions} />
+        <GoalsAndProjects space={space} goals={goals} projects={projects} toolsCount={2} />
+        <Discussions space={space} discussions={discussions} toolsCount={2} />
       </div>
```
Rockyy174, line 24: When we have 3 tools in a row, they are a bit squeezed and don't look that great. I've added `toolsCount` so that, when there are less than 3 tools, we don't squeeze them unnecessarily. 

Now we have the number 2 hardcoded, since we always have 2 tools, but when we start adding more tools, this number should change depending on how many tools the user is using.

PR #1443, opened 2024-10-29
chore: Update project check-in and goal update forms

Rockyy174:
I've updated the forms used to create and edit project check-ins and goal updates.

This PR also includes a migration which adds `status` to the goal updates schema. In the next PRs, I intend to update the goal update forms, so that the update status is also saved.


assets/js/pages/ProjectCheckInEditPage/Form.tsx
```diff
@@ -0,0 +1,95 @@
...
+function DescriptionSection({ checkIn }: { checkIn: ProjectCheckIn }) {
+  assertPresent(checkIn.project, "project must be present in checkIn");
+
+  const mentionSearchScope = { type: "project", id: checkIn.project.id! } as const;
+
+  return (
+    <div>
+      <div className="text-lg font-bold mb-2">2. What's new since the last check-in?</div>
+
+      <Forms.RichTextArea
```
shiroyasha, line 88: Can we use the label for this? (or is it visually not the same?)

assets/js/components/Forms/SelectStatus.tsx
```diff
@@ -0,0 +1,83 @@
...
+import * as Popover from "@radix-ui/react-popover";
+import * as People from "@/models/people";
+
+import { Status, Placeholder, StatusOptions } from "@/components/status";
+import { useFieldError, useFieldValue } from "./FormContext";
+
+interface Props {
+  field: string;
+  options: StatusOptions[];
+  reviewer?: People.Person;
```
shiroyasha, line 16: We should allow setting the label as well.

shiroyasha:
Nice 👏 This was a huge effort and a very good step forward to simplify our codebase.

assets/js/pages/ProjectCheckInEditPage/Form.tsx
Rockyy174: It looks almost identical. The only difference is the `mb-2` class which adds a space between the label and the element.

Here is what it's looks like without using the label (and how it currently is in production):
tmp-1

Here is with the label:
tmp-2

The same apply for project check-ins and goal updates forms. Do you think this is space between the elements is necessary?

shiroyasha: They look almost identical. Lets go with the one that is the default in forms to boost universality in the app.

PR #1464, opened 2024-10-31
feat: Goals & Projects tools shows pie chart summarizing resources status

Rockyy174:
The pie chart in the Goal & Projects tool is now built using goal updates status.

### DIFF
--- a/assets/js/components/PieChart/index.tsx
+++ b/assets/js/components/PieChart/index.tsx
@@ -2,7 +2,7 @@ import React from "react";
 
 interface Slice {
   size: number;
-  color: "green" | "yellow" | "red";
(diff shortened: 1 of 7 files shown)


assets/js/components/PieChart/index.tsx
```diff
@@ -12,9 +12,10 @@ interface PieChartProps {
 }
 
 const COLORS = {
-  green: "#10b981",
-  yellow: "#fde047",
-  red: "#ef4444",
+  green: "rgb(22, 163, 74)",
```
Rockyy174, line 15: I've changed the pie chart colors so that they are the same as the ones used in the status circle.

PR #1471, opened 2024-10-31
feat: Show update status on goal update page

Rockyy174:
The update status is now displayed on the goal update page:

tmp


shiroyasha:
Nice! I like the conditions rounded up with a border.

PR #1472, opened 2024-10-31
feat: Remove space navigation

Rockyy174:
Last time @shiroyasha and I talked about the Space Tools, it was decided that we would remove the space navigation, since we won't need it after we add the tools.

However, once I started working on it, I realized that, if we remove the navigation, when the user is on the Goals or Discussions page, they won't be able to navigate to other pages.

So, unless I'm missing something, the navigation should be removed only on the Space home page. So, I've implemented it this way, but it feels a bit weird to me when the navigation appears and disappears:

https://github.com/user-attachments/assets/f09d5574-1c36-456a-ab64-b5879b560252

Is this how it's supposed to look like? Or do you have any idea on how to improve it?

### DIFF
--- a/assets/js/components/SpacePageNavigation/index.tsx
+++ b/assets/js/components/SpacePageNavigation/index.tsx
@@ -15,9 +15,10 @@ import { assertPresent } from "@/utils/assertions";
 interface SpacePageNavigationProps {
   space: Spaces.Space;
   activeTab: "overview" | "discussions" | "goals" | "projects";
+  hideWhenOnOverview?: boolean;
 }
 
-export function SpacePageNavigation({ space, activeTab }: SpacePageNavigationProps) {
+export function SpacePageNavigation({ space, activeTab, hideWhenOnOverview }: SpacePageNavigationProps) {
   const { negTop, negHor } = Paper.usePaperSizeHelpers();
 
   const overviewPath = Paths.spacePath(space.id!);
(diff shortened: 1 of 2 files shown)


assets/js/components/SpacePageNavigation/index.tsx
```diff
@@ -15,9 +15,10 @@ import { assertPresent } from "@/utils/assertions";
 interface SpacePageNavigationProps {
   space: Spaces.Space;
   activeTab: "overview" | "discussions" | "goals" | "projects";
+  hideWhenOnOverview?: boolean;
 }
```
Rockyy174, line 19: The `hideWhenOnOverview` flag is necessary only because we are using the space navigation both on the original and temporary Space page. Once we have only one Space page, we can get rid of this flag.

shiroyasha:
Use this pattern on subpages (goals & project and discussions):

Screenshot 2024-10-31 at 22 08 18

The current project page implements a similar pattern. We can copy from there.

Rockyy174:
I've added a navigation similar to the one we have on the project subpages. This what it looks like now:

https://github.com/user-attachments/assets/6e6c47c6-a8bd-4917-b340-be7af4aa1809

shiroyasha (approved):
Looks good. Is it still a draft, or we are ready to 🚢 it?

Rockyy174:
A lot of tests are failing because of the change in the navigation. But as soon as I fix these tests, it should be ready.

PR #1474, opened 2024-11-01
feat: Replace old Goals page by new Goals & Projects page

Rockyy174:
I'm replacing the old Goals page by the new Goals & Projects page for company-wide goals. When the user visits a Space Goals page, they will still see the old goals tree.


Rockyy174:
I've added a new Button component which we can use in situations where there are several options:

https://github.com/user-attachments/assets/cdb0e19d-3e18-4865-87c9-b79992a8ba77

The cursor is actually "pointer", but when I record my screen, the cursor never changes.

I tried it with the round corners so that it would look like our other buttons, but it feels weird when there is a dropdown. 
@shiroyasha @markoa  Let me know if you think we can improve it!

shiroyasha:
I think it is good for the first pass. Once we use it live we will figure out the best way to optimize it.

PR #1480, opened 2024-11-04
feat: Show latest update/check-in on hover

Rockyy174:
The latest update/check-in is now displayed both when the user hovers over and clicks on the status. 

- When the user clicks on it, the popup is displayed until the user clicks outside of it. 
- When the user hovers over it for longer than a second, the popup is displayed until the user hovers outside of it for longer than a second.

https://github.com/user-attachments/assets/ce008b95-257e-4d77-9b6d-ec90b75bfdd8

Using `radix-ui/react-popover`, we got a nice popup that behaves nicely most of the time. However, it sometimes does some weird things, such as scrolling the window a little or changing the cursor from pointer to default. I spend some time trying to understand why it would sometimes behave like this, but couldn't find the reason. 

If it turns out that this behavior is not acceptable, I can spend some more time trying to fix. If there is no fix it, I can just build a popup from scratch.

### DIFF
--- a/assets/js/components/Buttons/OptionsButton.tsx
+++ b/assets/js/components/Buttons/OptionsButton.tsx
@@ -37,11 +37,12 @@ export function OptionsButton({ options, align = "center", testId }: Props) {
           sideOffset={5}
         >
           <Popover.Arrow className="fill-surface-outline scale-150" />
-          {options.map((option) => (
+          {options.map((option, idx) => (
             <div
               onClick={option.action}
               className="cursor-pointer px-4 py-2 border-b border-surface-outline hover:bg-surface-accent last:border-b-0"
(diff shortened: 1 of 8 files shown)


shiroyasha:
> If it turns out that this behavior is not acceptable, I can spend some more time trying to fix. If there is no fix it, I can just build a popup from scratch.

Looks good for first pass.

markoa:
It looks like you "hijacked" the click event and now it's impossible to open the check-in.

Rockyy174:
It seems I misunderstood you. Currently, when you click on "status", it shows the popup. The difference between clicking and hovering is that clicking on it will keep the popup open until the user manually closes it.

Would you like me to change it so that the user is redirected to the check-in page?

PR #1488, opened 2024-11-05
feat: Save Goals view options to localStorage

Rockyy174:
View options on Goals page are saved to local storage:

https://github.com/user-attachments/assets/ebd02e22-4277-454c-8164-7d3250b56225


test/support/features/goal_tree_steps.ex
```diff
@@ -64,6 +64,12 @@ defmodule Operately.Support.Features.GoalTreeSteps do
     UI.visit(ctx, Paths.goals_path(ctx.company))
   end
 
+  step :visit_goal_page, ctx, goal do
+    ctx
+    |> UI.visit(Paths.goal_path(ctx.company, goal))
+    |> UI.sleep(200)
```
shiroyasha, line 70: Instead of sleep we now have a better technique:

```
|> UI.assert_has(testid: "goal-page")
```

The tests id are added to every page in `<Pages.Page ... testId={...}>`.

Rockyy174: Got it!

PR #1491, opened 2024-11-05
feat: Align tree nodes vertically

Rockyy174:
Nodes in the goals tree are now aligned:

https://github.com/user-attachments/assets/cdc83410-8d1d-412c-ae00-ea3def41a42e


shiroyasha:
Much nicer! I like how everything is aligned now.

PR #1535, opened 2024-11-07
feat: Add resource hubs, nodes and folders schemas, and create resource hubs for spaces

Rockyy174:
I've added the schemas and migrations for the resource_hubs, nodes and folders. 

I've also added a data migrations which creates one resource hub for every existing space.


shiroyasha:
Wow this was fast. 👏

Issue #1531, opened 2024-11-07
Goals box in space grid: Remove status circles, color the progress bars


markoa:
This looks more straightforward:
<img width="397" alt="Screenshot 2024-11-07 at 13 05 15" src="https://github.com/user-attachments/assets/9a015c58-770b-4258-b6a2-09e1ac843b30">

Than this:
<img width="381" alt="Screenshot 2024-11-07 at 13 06 21" src="https://github.com/user-attachments/assets/ebb6d5f6-ed1c-4647-bcac-5643bd41e3dd">

PR #1537, opened 2024-11-07
feat: Simplify Goals & Projects tool

Rockyy174:
Addresses https://github.com/operately/operately/issues/1531. 

I noticed that with this solution, we lost some information. There is no way to know if a goal is in `caution` or `issue` state, if it has no progress yet.

For example:

Before:
tmp

Now:
tmp2

Previously, we could see that the 2nd goal was in `caution` state, but now we can't. 
Maybe that's not very important, but I thought it was worth mentioning it.


assets/js/components/charts/constants.tsx
```diff
@@ -0,0 +1,6 @@
+export const COLORS = {
```
Rockyy174, line 1: I've grouped some components in the `charts` directory because I wanted to use the same colors in them, but found it messy to import it from a different component folder. Now the `COLORS` used in the components lives in `charts/constants.tsx`

shiroyasha:
Good point about the statuses. Let's sleep on it before we eliminate a status indicator.

markoa:
I still prefer the reduced option...I understand the concern about not being 100% clear about non-on-track statuses but the box is supposed to be a preview not a comprehensive dashboard.

shiroyasha:
Screenshot 2024-11-13 at 19 18 25

First time seeing this message.

Rockyy174:
@shiroyasha It should be good now.

Issue #666, opened 2024-07-02
Adding / removing a (project) resource doesn't create a feed event


markoa:
- open a project
- click "edit" within key resources
- add one
- feed is unchanged, should mention creation of the asset
- same after removing

markoa:
@Rockyy174 Since you added various missing feed events recently, can you verify if this has been addressed?

Rockyy174:
Sure!

PR #1543, opened 2024-11-08
feat: Adding a key resource shows event in the feed

Rockyy174:
Handles the "adding key resource" part of https://github.com/operately/operately/issues/666.


shiroyasha:
Heh issue #666.

Issue #1530, opened 2024-11-07
Make Goals & Projects responsive


markoa:
probable direction:
- lower font size a little bit on small screens
- for projects, break before milestonestones into a new row
- gracefully handle avatar lists
<img width="489" alt="Screenshot 2024-11-07 at 12 49 29" src="https://github.com/user-attachments/assets/70caa13b-150f-4532-be09-04b10c569906">

This screen is the heart of the system so imo definitely worth it not delaying much.

Issue #1527, opened 2024-11-07
Visually distracting pop-in/pop-out on Goal Map


shiroyasha:
Gives me the same feeling as Notion, wherever I go something is hoping in and on the page. Forces me to focus on that instead of what I'm reading.

I see several ways to fix it:

a) Introduce a ... on the right side which is always visible
b) Have always present buttons/icons that dim-in/dim-out as you hover the row
c) Maybe just moving these buttons completely on the right can fix it as it would not just in at what my brain conceives as "random" locations, but in a consistent place.

https://github.com/user-attachments/assets/8869b7fb-0604-402f-9c33-54a0e6a6cec2

markoa:
Addressed in the upcoming Work Map, technically #2040.

PR #1546, opened 2024-11-08
feat: Goals page is responsive

Rockyy174:
Fixes https://github.com/operately/operately/issues/1530.

Regarding decreasing the font-size, it didn't seem to make a big difference and, in the resource details, the font is already pretty small.

Something that still doesn't look good in my opinion is the "add sub-goal" and "create project" buttons. I'm not really sure what to do with them when we run out of space. Is it an option to hide them when the screen is in the "xs" breakpoint?

https://github.com/user-attachments/assets/9cb07b66-fc08-4c7c-aecb-eaf223724411

### DIFF
--- a/assets/js/components/ProgressBar/index.tsx
+++ b/assets/js/components/ProgressBar/index.tsx
@@ -1,8 +1,14 @@
 import React from "react";
 import classNames from "classnames";
 
-export function ProgressBar({ percentage, className }: { percentage: number; className?: string }) {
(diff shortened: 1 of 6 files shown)


shiroyasha:
Let's go with hiding them on xs. It is unrealistic that someone would define goals on a mobile withe current flow.

We might reintroduce it later depending on the outcome of https://github.com/operately/operately/issues/1527.

PR #1547, opened 2024-11-11
fix: Move action buttons to the right on Goals page

Rockyy174:
I've moved the action buttons to the right as suggested in https://github.com/operately/operately/issues/1527. 

https://github.com/user-attachments/assets/c1acabbe-4e95-4fb4-8e70-3006d75586c4


shiroyasha:
I'm not yet sure if this will be the final solution, but let's try it out.

PR #1561, opened 2024-11-13
feat: Add MessagesBoard

Rockyy174:
I've added the `Operately.Messages.MessagesBoard` schema.

Originally, I wanted to finish the Resource Hubs before working on the other Space Tools, but then I realized that I would have to add a query to fetch the resource hubs to the frontend. 

If I wrote a `ListResourceHubs` query now, it would be a waste of time, since in the near future I would replace it by a `ListSpaceTools` query. 

So, I decided to add the `MessagesBoard` and `GoalsAndProjectsGroup` schemas now. Then, I will add the `ListSpaceTools` query which will fetch all the existing tools in a Space. 

It will be a lot of work now, but later, the backend will be mostly ready for handling the Space Tools.

### DIFF
--- a/lib/operately/groups/insert_group.ex
+++ b/lib/operately/groups/insert_group.ex
@@ -6,10 +6,10 @@ defmodule Operately.Groups.InsertGroup do
   def insert(multi, attrs) do
     multi
     |> insert_group(attrs)
-    |> insert_group_context()
-    |> insert_members_access_group()
-    |> insert_managers_access_group()
+    |> insert_context()
+    |> insert_access_groups()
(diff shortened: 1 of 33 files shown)


lib/operately/groups/insert_group.ex
```diff
@@ -66,4 +62,14 @@ defmodule Operately.Groups.InsertGroup do
       end
     end)
   end
+
+  defp insert_default_messages_board(multi) do
+    multi
```
Rockyy174, line 67: While we don't have a way to manually add message boards, it's necessary to create a default one with every space.

PR #1584, opened 2024-11-15
feat: Folders can be created and listed in the Resource Hub 

Rockyy174:
Now, it's possible to create folders in the Resource Hub:

https://github.com/user-attachments/assets/2afc386d-5b92-4ce1-8bce-e8519ba7dbb2


assets/js/pages/SpaceResourceHubPage/page.tsx
```diff
@@ -36,29 +44,10 @@ function PageNavigation() {
...
-        options={[
-          { icon: IconFile, label: "Write a new document", action: () => {}, testId: "new-document" },
-          { icon: IconFolder, label: "Create a new folder", action: () => {}, testId: "new-folder" },
-          { icon: IconUpload, label: "Upload files", action: () => {}, testId: "upload-files" },
-        ]}
-        testId="add-options"
-      />
-    </div>
-  );
+  return <div className="text-content-accent text-center text-3xl font-extrabold">{resourceHub.name}</div>;
```
shiroyasha, line 50: We have a `<Paper.Title>` component. Can we use that one?

I want to us to slowly switch to a shared component for titles that will keep the visual look&feel of the application consistent across the application.

lib/operately_web/api/queries/get_resource_hub.ex
```diff
@@ -0,0 +1,32 @@
+defmodule OperatelyWeb.Api.Queries.GetResourceHub do
+  use TurboConnect.Query
+  use OperatelyWeb.Api.Helpers
+
+  alias Operately.ResourceHubs.ResourceHub
+
+  inputs do
+    field :id, :string
```
shiroyasha, line 8: Can we use the `:id` type? That way we don't need to manually decode it, which results in less code and a uniform behaviors across the API.

assets/js/pages/SpaceResourceHubPage/page.tsx
Rockyy174: Sure! I didn't know about that component.

lib/operately_web/api/queries/get_resource_hub.ex
Rockyy174: Cool, I didn't know about that either.

shiroyasha:
Ah, sorry. I made a mistake, that is an old (maybe even unused component). 
The correct component is `<Paper.Header>`.

Here is one examples of its usage:

- [Discussions Page](https://github.com/operately/operately/blob/main/assets/js/pages/SpaceDiscussionsPage/page.tsx#L40) with title-center-actions-left layout

- [Company Amin page](https://github.com/operately/operately/blob/main/assets/js/pages/CompanyAdminManageAdminsPage/page.tsx#L32-L35) with title-left-actions-right layout

PR #1601, opened 2024-11-20
feat: Add ResourceHubFolderPage page

Rockyy174:
I've added the `ResourceHubFolderPage` page.


assets/js/pages/ResourceHubNewDocumentPage/form.tsx
```diff
@@ -29,8 +30,12 @@ export function Form() {
       navigate(Paths.resourceHubPath(resourceHub.id!));
     },
     submit: async () => {
+      const params = new URLSearchParams(location.search);
+      const folderId = params.get("folderId");
```
shiroyasha, line 34: Let's move all url search params lookup to the page loader to isolate components from the page as much as possible.

PR #1679, opened 2024-12-04
chore: Add vite to project

Rockyy174:
This is an attempt to add Vite to the project.


shiroyasha:
Vite was introduced in https://github.com/operately/operately/pull/2235

Issue #1732, opened 2024-12-18
Docs: More engaging zero-state copy


markoa:
Change from current "Nothing here..." to

**Ready for your first document**

Your team's central hub for sharing documents, images, videos, and files. Click 'Add' to get started.

(add one line of empty space between)

PR #1735, opened 2024-12-18
chore: Improve resource hub zero-state copy

Rockyy174:
Fixes https://github.com/operately/operately/issues/1732.

This is how the page looks like now:

tmp


shiroyasha:
Will this be displayed only if the resource hub is empty, or in every empty folder?

Rockyy174:
I've just checked it, @shiroyasha. It's displayed in empty folders. I suppose that the new copy doesn't make much sense for folders. Should we keep using the previous one for folders? Or will we come up with something better for folders, too?

shiroyasha:
Let's use this one only if repohub is completely empty, while for emtpy folders let's raise up this question the issue.

markoa:
(same title)
This folder is empty. Click 'Add' to upload your first file.

shiroyasha (approved):

PR #1736, opened 2024-12-18
chore: Improve drag-and-drop animation

Rockyy174:
I've changed the files drag-and-drop animation. This is what it looks like now:

https://github.com/user-attachments/assets/de51f8dd-8217-43f4-8dde-7afb16f38b4e


shiroyasha:
Nice! Will it also look ok in dark mode? Asking because I noticed some classes that might be problematic.

Rockyy174:
Unless I missed some detail, it should be ok in the dark mode, too:

https://github.com/user-attachments/assets/13c348c4-0779-4b78-9acd-c32a9c9eddee

Issue #1745, opened 2024-12-18
Docs & Files: Indicate presence of comment(s)


markoa:
In the file list, a simple number-in-a-circle on the right hand side showing number of comments (when non-zero) will be a useful indicator of activity.

PR #1758, opened 2024-12-19
feat: Display comments count for documents and files

Rockyy174:
Fixes https://github.com/operately/operately/issues/1745.

The logic for loading the comments count is in place. Since comments count is something we will have to display in many places, I made a component for it.

I'm just not sure where it looks better. Here is what it looks like right after the description:
tmp-1

And here is what it looks like at the end of the container:
tmp-2

If we decide to go with the second picture, then this PR should be ready. If the first picture is better, then I would have to remove the last commit.

Also, if you think the count would look better positioned somewhere else, just let me know!

### DIFF
--- a/assets/js/api/index.tsx
+++ b/assets/js/api/index.tsx
@@ -1100,7 +1100,7 @@ export interface ResourceHubDocument {
   insertedAt?: string | null;
   permissions?: ResourceHubPermissions | null;
   reactions?: Reaction[] | null;
-  comments?: Comment[] | null;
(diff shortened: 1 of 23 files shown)


shiroyasha:
The second form will be a good base for further (planned) UI improvements.
Please add some feature tests that verify if the comment count is visible.

shiroyasha (changes requested):
The second form will be a good base for further (planned) UI improvements.
Please add some feature tests that verify if the comment count is visible.

PR #1762, opened 2024-12-20
feat: Add ResourceHubCopyDocumentPage

Rockyy174:
I've added the "copy" action to documents:

https://github.com/user-attachments/assets/aa428461-bcca-416d-9f2e-e7707a1f16c4

This is what it looks like in the feed:
tmp


shiroyasha (changes requested):
Everything looks good, but please add feature tests.

PR #1776, opened 2024-12-26
feat: Links can be created

Rockyy174:
Links can now be added to a Resource Hub. However, we cannot add specific links yet, such as a Google Sheet or Doc link. I will handle those specific links in the near future.


Rockyy174:
The bot suggested that we add URL validation. I didn't think about that since users will be adding URLs for themselves. Do you think it's necessary for our use case,  @shiroyasha?

shiroyasha:
I think it is worth adding to check basic stuff like: Does it start with a `http://` or `https://`, sometimes copy paste issues creep in, let's protect our users from it.

shiroyasha (approved):

PR #1801, opened 2024-12-30
fix: Document navigation includes complete path back to the space

Rockyy174:
The navigation in the Document page now includes to all the document's nested parent folders, resource hub and space:

https://github.com/user-attachments/assets/dd37639b-e962-442f-9035-441ee6a1f528

### DIFF
--- a/lib/operately/resource_hubs/node.ex
+++ b/lib/operately/resource_hubs/node.ex
@@ -72,4 +72,48 @@ defmodule Operately.ResourceHubs.Node do
       end
     end)
   end
+
+  @doc """
+  Takes a resource ID and the name of the resource table and
+  recursively queries all the parent folders of the resource.
+
+  Example:
+    find_all_parent_folders(folder.id, "resource_folders")
+    find_all_parent_folders(document.id, "resource_documents")
+  """
+  def find_all_parent_folders(resource_id, resource_table) do
+    allowed_tables = ["resource_folders", "resource_documents", "resource_links"]
+
+    unless resource_table in allowed_tables do
+      raise ArgumentError, "#{resource_table} is not an allowed table. Allowed tables: #{allowed_tables |> Enum.join(", ")}"
+    end
+
+    q = """
+      WITH RECURSIVE folder_hierarchy AS (
+        SELECT r.id, n.parent_folder_id, n.name
+        FROM #{resource_table} r
+        JOIN resource_nodes n ON r.node_id = n.id
+        WHERE r.id = $1
+
+        UNION ALL
+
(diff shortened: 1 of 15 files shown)


lib/operately/resource_hubs/node.ex
```diff
@@ -72,4 +72,48 @@ defmodule Operately.ResourceHubs.Node do
...
+        FROM resource_folders f
+        JOIN resource_nodes n ON f.node_id = n.id
+        JOIN folder_hierarchy fh ON f.id = fh.parent_folder_id
+
+      )
+      SELECT id, name FROM folder_hierarchy WHERE id != $1;
+    """
+
+    {:ok, id} = Ecto.UUID.dump(resource_id)
+    {:ok, %{rows: rows}} = Operately.Repo.query(q, [id])
```
shiroyasha, line 110: Can we pass the table name as a parameter here? Should be safer then interpolating.

test/features/resource_hub_test.exs
```diff
@@ -210,4 +197,32 @@ defmodule Features.Features.ResourceHubTest do
...
+      |> Steps.refute_navigation_links(["three", "four"])
+      |> Steps.assert_navigation_links(["Resource hub", "one", "two"])
+      |> Steps.navigate_back("one")
+      |> Steps.refute_navigation_links(["one", "two"])
+      |> Steps.assert_navigation_links(["Resource hub"])
+    end
+
+    feature "document navigation works", ctx do
+      ctx
+      |> Steps.give_document_within_nested_folders_exists()
```
shiroyasha, line 217: *given

lib/operately/resource_hubs/node.ex
Rockyy174: Yeah, it should. Actually, I tried it, but it threw an error. What I did was something like the following:

```
    q = """
      WITH RECURSIVE folder_hierarchy AS (
        SELECT r.id, n.parent_folder_id, n.name
        FROM $2 r
        JOIN resource_nodes n ON r.node_id = n.id
        WHERE r.id = $1
        UNION ALL
        SELECT f.id, n.parent_folder_id, n.name
        FROM resource_folders f
        JOIN resource_nodes n ON f.node_id = n.id
        JOIN folder_hierarchy fh ON f.id = fh.parent_folder_id
      )
      SELECT id, name FROM folder_hierarchy WHERE id != $1;
    """

    {:ok, %{rows: rows}} = Operately.Repo.query(q, [id, resource_table])
```

Do you know any other way of doing this?

shiroyasha: I thought that would be it. If it doesn't work that way, let's merge as it is.

PR #1807, opened 2024-12-30
docs: New space form, remove 'team' from form tip

markoa:
Tiny detail but it would be wrong to lead people to think that it is optimal to have " Team" appended to every space name.


shiroyasha:
@markoa it seems that you don't have your git configured. Here's how to do it: https://github.com/operately/operately/blob/main/docs/commit_sign-off.md

markoa:
I comitted straight in GitHub web. Fixed the DCA thing just now.

Issue #1576, opened 2024-11-15
Cannot post discussion if large file is uploaded


AleksandarCole:
**User story:**
As a user I want to post a new discussion to update the team of new feature, in this discussion I want to attach video showcasing the feature. 

**Bug:**
I can create the discussion and drag&drop video that seems uploaded (renders and plays in edit mode). I write/add the rest of the discussion. When I try to post it, nothing happens. I click the button multiple times, resulting eventually in 500. My whole progress is wiped and I have to write post from scratch. I also don't know root cause (video file too large), causing me to repeat the issue once more. 

**Steps to reproduce:** 
- Create a new discussion in any space
- Drag&Drop the video attached at the bottom of this issue
- Wait for video to fully upload (green progress bar)
- Add some text
- Try to post. Button won't react. 
- Refresh page (text is gone)

Note - I was not able to reproduce getting 500 though, although it happened first time I got this bug.

**Root cause**
Unhandled `413 - payload too large` error
![image](https://github.com/user-attachments/assets/7980c427-4c08-45cf-ad34-648e55934895)

**Proposed solutions**
- Give user feedback on file upload (if it's too large, block it right away)
- OR: Give user error when they try to post
- AND: Give users options to have `Draft` posts where content is saved without posting it

shiroyasha:
Thanks for the detailed brakedown. It helps a ton.
I've edited the original message to break down the problem into smaller manageable chunks.

shiroyasha:
One of the steps to fix this issue [draft messages](https://github.com/operately/operately/issues/692) is released.

markoa:
This happens when editing documents too. Just experienced this while trying to add a 2.7MB image to an existing doc.

shiroyasha:
@markoa was it a clean-shot x screenshot? Context: It seems that I'm able to reproduce this behavior on smallish files (2mb) only if the filename contains an `@` symbol. The same with pasting here: https://github.com/operately/operately/issues/999.

markoa:
I confirm there is no issue either pasting or uploading large images inside a doc, with or without funny characters.

PR #1830, opened 2025-01-04
fix: Handle large files

Rockyy174:
We were storing the size of files as integers in the `blobs` schema. However, as we were storing the size of files as its number of bytes, a large file would have too many bytes and wouldn't fit in the Postgres integer field. After modifying the size field from integer to bigint, uploads of large files started working.

I'm not sure if these changes will be enough to fix the issue in production, but it was enough to fix it locally.


shiroyasha (approved):

shiroyasha:
related to https://github.com/operately/operately/issues/1576.

PR #1848, opened 2025-01-09
chore: CreateBlob mutation can be used to create multiple blobs

Rockyy174:
We can now use the `OperatelyWeb.Api.Mutations.CreateBlob` mutation to create multiple blobs at once.


shiroyasha (approved):
Looks good! 🚀

PR #1873, opened 2025-01-16
feat: Folder can be copied

Rockyy174:
Folders can now be copied.

### DIFF
--- a/assets/js/features/ResourceHub/components/CopyFolder.tsx
+++ b/assets/js/features/ResourceHub/components/CopyFolder.tsx
@@ -33,9 +33,15 @@ export function CopyFolderModal(props: FormProps) {
         type: "pathToFolder" in parent ? "folder" : "resourceHub",
       },
     },
+    validate: (addError) => {
+      if (!form.values.name) {
+        addError("name", "Name is required");
+      }
+    },
     cancel: hideModal,
     submit: async () => {
       const res = await post({
+        folderName: form.values.name,
         folderId: resource.id,
         destResourceHubId: resource.resourceHubId,
         destParentFolderId: form.values.location.type == "folder" ? form.values.location.id : undefined,
--- a/assets/js/features/ResourceHub/components/CopyFolder.tsx
+++ b/assets/js/features/ResourceHub/components/CopyFolder.tsx
@@ -33,11 +33,6 @@ export function CopyFolderModal(props: FormProps) {
         type: "pathToFolder" in parent ? "folder" : "resourceHub",
       },
     },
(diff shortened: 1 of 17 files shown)


assets/js/features/ResourceHub/components/CopyFolder.tsx
```diff
@@ -0,0 +1,55 @@
...
+    fields: {
+      name: resource.name + " - Copy",
+      location: {
+        id: parent?.id,
+        type: "pathToFolder" in parent ? "folder" : "resourceHub",
+      },
+    },
+    validate: (addError) => {
+      if (!form.values.name) {
+        addError("name", "Name is required");
```
shiroyasha, line 38: There is a `required` prop that you can add to the input field. Can we use that one?

Rockyy174: Sure!

assets/js/features/ResourceHub/components/CopyFolder.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const navigate = useNavigate();
+
+  assertPresent(resource.resourceHubId, "resourceHubId must be present in resource");
+
+  const form = Forms.useForm({
+    fields: {
+      name: resource.name + " - Copy",
+      location: {
+        id: parent?.id,
+        type: "pathToFolder" in parent ? "folder" : "resourceHub",
```
shiroyasha, line 33: I've noticed this pattern the other day I was working with resources. It is clean now, but it won't be clear in a couple of weeks when we come back to this area.

I see two solutions:

1. we somehow add a `type` to the resources
2. or we make an a utility function somewhere in the resource hub model and document the behaviour.

For (2) we could do something like:

```
function findTypeOfNodeParent(parent) : "folder" | "resourceHub" {
  // figure out if the node is a folder or resourceHub
  return "pathToFolder" in parent ? "folder" : "resourceHub";
}
```

This is still a hack, but at least it has a bit more context.
Ideally, we would go with (1).

---

One thing that is very nice in GraphQL is that every API object contains the name of the type in the JSON response under `__type__`. For example, in our case it would be `parent.__type__`. Maybe we could generalize this for all objects returned from the API?

shiroyasha: Yeah, me too, but I'm not sure how achievable it is.

Let's do the following:
1. merge this PR as is
2. pls open an issue for adding this type with dev-productivity label
3. invest some time next week to estimate if it feasable

PR #1893, opened 2025-01-20
feat: Documents can be saved as drafts

Rockyy174:
Users can now save documents as drafts.


assets/js/components/Forms/useForm/state.tsx
```diff
@@ -1,6 +1,6 @@
 import React from "react";
 
-export type State = "idle" | "uploading" | "validating" | "submitting";
+export type State = "idle" | "uploading" | "validating" | "submitting" | "secondary-submitting";
```
shiroyasha, line 3: Instead of having `submitting` and `secondary-submitting` maybe it would be better to allow passing additional data to the submit function via the buttons.

Something like:

```
<Forms.SubmitButton onClick={() => forms.submit("as-draft")}>
<Forms.SubmitButton onClick={() => forms.submit("publish")}>
```

Would be more explicit and could handle more varying use cases that we encounter in the future.

Rockyy174: Okay. I will try to do it this way.

assets/js/components/Forms/useForm.tsx
```diff
@@ -54,7 +59,7 @@ export function useForm<T extends FieldObject>(props: FormProps<T>): FormState<T
       removeValidation,
       getValue,
       setValue,
-      submit: async () => {
+      submit: async (attrs: any) => {
```
shiroyasha, line 62: Should this be `attrs?: any`?

shiroyasha:
Looks nice! I definitely like it better than the extra state in the form's state machine. Shipit! 🚢

PR #1906, opened 2025-01-22
feat: Resource Hub notifications are marked as read when page loads

Rockyy174:
When a Resource Hub resource page loads, unread notifications related to that resource are marked as read.

I've added `Operately.Notifications.UnreadNotificationsLoader` and was able to remove a lot of repeated code.


shiroyasha:
Nice improvement for the codebase!

PR #1915, opened 2025-01-23
fix: Stabilize UI.refute_text/3

Rockyy174:
This is an attempt to stabilize `UI.refute_text/3` which got kind of flaky since the Resource Hub tests were added.

### DIFF
--- a/test/support/features/ui.ex
+++ b/test/support/features/ui.ex
@@ -2,7 +2,7 @@ defmodule Operately.Support.Features.UI do
   use ExUnit.CaseTemplate
   alias Wallaby.Query
 
-  alias Wallaby.Browser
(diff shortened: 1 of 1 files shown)


Rockyy174:
Our `refute_text(state, text, testid: id)` function was taking too long to execute and was timing out, that's why we started having some flaky tests lately. 

This function was using `Browser.refute_has(Query.text(text))`. If I'm not mistaken, this function already caused us problems with too long execution time in `refute_text/2` some months ago 🤔 But I am not sure.

So, I rewrote `refute_text/3` without using `Browser.refute_has`. It seems to have worked. Let me know if you think these changes are okay, @shiroyasha

Issue #1917, opened 2025-01-23
Changing parent goal doesn't create an activity


markoa:
I changed a goal's parent goal owned by somebody else (company curator mode), but the change doesn't appear in the goal's activity feed. I guess that the owner hasn't received a notification then either.

PR #1926, opened 2025-01-24
fix: GoalReparent event appears in the feed and sends notifications

Rockyy174:
Fixes https://github.com/operately/operately/issues/1917.

This was a tricky problem. Currently, `GoalReparent` activities only have the `company`, `old_parent_goal` and `new_parent_goal` in its content, but they lack a reference to the specific goal that was reparented. This makes it impossible to determine which goal the activity belongs to.

When there is missing data in an activity, the solution is usually to run a data migration to add/create the missing data. However, in this case, I didn't see any way to find out which goal the activity belonged to.

Another problem is that currently we don't have a clean way to filter out activities that have missing data. So, the activities with missing data were breaking the feed. 
Of course, I could handle those activities so that they don't break the feed, but then the feed item would be pretty much meaningless. 

Therefore, I decided to write a migration to delete these activities. 
I'm aware that any other solution would be better than deleting the activities, but I couldn't come up with any, considering that I couldn't fix the missing data in these existing activities and they serve no purpose without their associated goal references.

### DIFF
--- a/assets/js/api/index.tsx
+++ b/assets/js/api/index.tsx
@@ -300,9 +300,9 @@ export interface ActivityContentGoalReopening {
 }
 
 export interface ActivityContentGoalReparent {
-  companyId?: string | null;
-  oldParentGoalId?: string | null;
-  newParentGoalId?: string | null;
+  goal?: Goal | null;
+  oldParentGoal?: Goal | null;
+  newParentGoal?: Goal | null;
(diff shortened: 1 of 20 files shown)


shiroyasha:
Good solution! I'll merge it later when I can follow through the deployment process just in case some problems surface up.

Issue #1897, opened 2025-01-21
New goal reviewer list doesn't take into space access levels


markoa:
Drastic example: set up a private space with just yourself. Click new goal, and the list will offer everyone in the company.

PR #1930, opened 2025-01-27
fix: Only show possible contributors who have access when creating a goal

Rockyy174:
Fixes https://github.com/operately/operately/issues/1897.

### DIFF
--- a/assets/js/features/goals/GoalForm/Form.tsx
+++ b/assets/js/features/goals/GoalForm/Form.tsx
@@ -3,6 +3,8 @@ import React from "react";
 import classnames from "classnames";
 import PeopleSearch from "@/components/PeopleSearch";
 
+import { Space } from "@/models/spaces";
+import { Goal } from "@/models/goals";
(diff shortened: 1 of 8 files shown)


assets/js/features/goals/GoalForm/Form.tsx
```diff
@@ -234,3 +240,13 @@ function SectionHeader({ form, title, subtitle }: { form: FormState; title: stri
     </div>
   );
 }
+
+function findScope(space: Space | null, goal: Goal | undefined) {
+  if (goal) {
+    return People.goalScope(goal.id!);
```
shiroyasha, line 246: Not sure about this one, it means that once you have a goal you can only select people who are already on the goal?

Rockyy174: Not exactly. You will be able to select anyone who has access to the goal. That would include space members and even company members, but only the ones who can see the goal. 

However, changing it to always rely on who is part of the space would be very easy. Just let me know if I should make these changes!

Issue #1918, opened 2025-01-23
Sluggish comment posting


markoa:
https://github.com/user-attachments/assets/84b304f7-bb77-4757-83c2-d4d0b2069458

For a while there is a period when both the new comment and the old comment are present on screen -- in two phases:

![Image](https://github.com/user-attachments/assets/502ef4d8-4014-44c6-83cb-0e61f498b04d)

![Image](https://github.com/user-attachments/assets/77a896c0-45f5-4432-acfe-c25a4ff8c402)

this is impossible to not notice and gives a "low quality" vibe tbh.

markoa:
Just to clarify, this is not temporary or new. I see it everyday across localhost, self-hosted and saas.

PR #1943, opened 2025-01-29
fix: Improve inconsistent behavior in comments

Rockyy174:
Fixes https://github.com/operately/operately/issues/1918.

This PR addresses several issues that were causing inconsistent behavior in the comment functionality. Hopefully, all the problems related to comments have now been resolved.

Also, I've consolidated all the comment logic into hooks that are now used for all the commentable resources. So, the improvements should be present consistently on every page with comment functionality.


lib/operately_web/api/mutations/post_milestone_comment.ex
```diff
@@ -47,6 +47,8 @@ defmodule OperatelyWeb.Api.Mutations.PostMilestoneComment do
       %{
         content: %{"message" => message},
         author_id: person.id,
```
Rockyy174, line 49: Comments added to a milestone only had a reference to the `MilestoneComment`, but not to the milestone itself. This caused a bug in the signal that is triggered when the comment is edited:
https://github.com/operately/operately/blob/d1cfecd95fcfec09a803f2466bd5bd957bf56033/lib/operately/operations/comment_editing.ex#L15C1-L21C8

As the comment didn't have a reference to the milestone stored in the `entity_id`, the signal wouldn't work, and the page wouldn't update when a comment is edited.

Now, new comments will have this reference to the milestone, fixing the but.

lib/operately_web/api/mutations/post_milestone_comment.ex
```diff
@@ -47,6 +47,8 @@ defmodule OperatelyWeb.Api.Mutations.PostMilestoneComment do
...
+        entity_id: milestone.id,
+        entity_type: :project_milestone,
       }
```
Rockyy174, line 52: Comments added to a milestone only had a reference to the `MilestoneComment`, but not to the milestone itself. This caused a bug in the signal that is triggered when the comment is edited:
https://github.com/operately/operately/blob/d1cfecd95fcfec09a803f2466bd5bd957bf56033/lib/operately/operations/comment_editing.ex#L15C1-L21C8

As the comment didn't have a reference to the milestone stored in the `entity_id`, the signal wouldn't work, and the page wouldn't update when a comment is edited.

Now, new comments will have this reference to the milestone, fixing the but.

shiroyasha (approved):
Looks great!

PR #1951, opened 2025-01-30
chore: Add Operately.Assignments.Loader

Rockyy174:
I've added a new module for loading a person's assignments. The `GetAssignments` query has already been updated to use the new loader.

In my next PR, I plan to update `OperatelyEmail.Emails.AssignmentsEmail` so that the emails and the Review page use the same data.

It's worth noting that the emails include due milestones, while the Review page does not. I've ensured that milestones won't break the page, but we might want to either display due milestones on the Review page or remove them from the emails to maintain consistency between the emails and the Review page.

### DIFF
--- a/assets/js/api/index.tsx
+++ b/assets/js/api/index.tsx
@@ -1266,12 +1266,13 @@ export interface ResourceHubUploadedFile {
 }
 
 export interface ReviewAssignment {
-  id?: string | null;
+  resourceId?: string | null;
   name?: string | null;
   due?: string | null;
   type?: string | null;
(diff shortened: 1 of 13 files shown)


shiroyasha:
Let's remove the milestones completly.

shiroyasha (dismissed):
The code looks good. Make sure to rebase before merging, I've done some small modifications to the review page that might break.

Rockyy174:
Sure, I will remove the milestones before merging it. As for rebasing, I think I already did.

Issue #1912, opened 2025-01-23
No goals lineage on profile page


markoa:
- open someone's profile (eg. Adriano in prod)
- open Goals tab
- only projects are listed

Context: trying to figure out "what is person X currently working on" brought me here.

Rockyy174:
I was investigating this bug, but I don't think I understood it. I went to the Goal's tab and I the goal tree shows both goals and projects. Am I looking in the wrong page?

https://github.com/user-attachments/assets/357c3931-19d4-4574-bf90-8a6871e651c1

shiroyasha:
I've discussed this issue with @Rockyy174 live. We will be showing the full ancestry of every project that is assigned to a person. 

Example: You are the champion of Project A, that has Goal A as parent.

Previously we would only show:

```
Project A
```

While with the "full lineage" we would show:

```
Goal A
  Project A
```

If Goal A has ancestry, we would show the full lineage up to the top:

```
Goal X
  Goal B
    Goal C
      Goal A
        Project A
```

PR #1965, opened 2025-02-05
fix: Goals tree on profile page includes complete goal lineage

Rockyy174:
Fixes https://github.com/operately/operately/issues/1912.


shiroyasha:
Looks good! 🚢

PR #1990, opened 2025-02-10
chore: Show error message to user if file upload fails

Rockyy174:
An error message is now displayed to the user if a file upload fails. 

I've added a very general error message as it catches any error it may occur.


shiroyasha:
Closing, stale pull request.

Issue #2020, opened 2025-02-15
Self-Hosted installation script warning before migrations


shiroyasha:
They are harmless, but they are introducing unnecessary noise and reduce the polished feeling of Operately.

![Image](https://github.com/user-attachments/assets/84df5bb7-9f27-437b-a5cc-32167e3ca1b7)

Issue #2019, opened 2025-02-15
Log In doesn't work on a fresh self-hosted insallation


shiroyasha:
Reported:

> I have followed the self-host guide and got the app up and running as expected. however, after I sign out, I was not able to sign back in. the sign in page doesn't show the login form. please see below screenshot.

![Image](https://github.com/user-attachments/assets/fb981863-9266-4229-b5f5-f1693a52fbe6)

PR #2022, opened 2025-02-15
fix: Login with email page should use allowLoginWithEmail instead of the allowLoginWithGoogle flag

shiroyasha:
fixes https://github.com/operately/operately/issues/2019
fixes https://github.com/operately/operately/issues/2020

### DIFF
--- a/assets/js/components/RichContent/Summary.tsx
+++ b/assets/js/components/RichContent/Summary.tsx
@@ -42,8 +42,6 @@ export function summarize(node: any): any {
       return summarizeOrderedList(node);
     case "mention":
       return summarizeMention(node);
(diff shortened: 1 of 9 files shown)


assets/js/components/RichContent/Summary.tsx
```diff
@@ -42,8 +42,6 @@ export function summarize(node: any): any {
       return summarizeOrderedList(node);
     case "mention":
       return summarizeMention(node);
-    case "heading":
-      return summarizeText(node);
```
shiroyasha, line 46: Unrelated to this change, just noticed in the JS build script that we had a warning for case duplication.

PR #2068, opened 2025-03-05
feat: Add new Goal Check-in page

Rockyy174:
I've added the new Goal Check-in page, which works for both viewing and editing a check-in.

https://github.com/user-attachments/assets/f4dddf77-cba0-43e4-b4bf-21c039704e40


shiroyasha:
Looks good! 🚀

PR #2075, opened 2025-03-06
feat: Display timeframe change in goal check-in feed event

Rockyy174:
If a Goal check-in includes a timeframe change, it will now be displayed in the feed. 

The timeframe changes in the `GoalCheckIn` and `GoalTimeframeEditing` activities now share the same component:

tmp


shiroyasha:
Visually we will need to make changes, but it is a good start to show it. 👍

PR #2122, opened 2025-03-17
feat: Goal and CheckIn pages share the same Targets field

Rockyy174:
The new Goal Page and the Goal Check-in Page now share the same Targets field. 

The borders which look nice in the Check-in Page didn't look good in the Goal Page:

I tried hiding the borders, but it became difficult to determine which values belonged to which targets:

Finally, I added some dots to connect targets and values:

In my opinion, the layout with the dots looks better than the others, though we can probably improve it further.

### DIFF
--- a/assets/js/components/Forms/GoalTargetsField.tsx
+++ b/assets/js/components/Forms/GoalTargetsField.tsx
@@ -1,159 +1,174 @@
 import React from "react";
 
+import { IconChevronDown } from "@tabler/icons-react";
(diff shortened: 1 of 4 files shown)


shiroyasha:
I agree, the third one looks the best + it works similarly to another element down on the page. Lets go with that one.
 
Note if we want to revert to the first one: To make the first one work, we could remove other separators on the page which are causing the problem atm.

PR #2197, opened 2025-03-27
chore: Increase parallelism for feature tests

shiroyasha:


shiroyasha:
Closing. Did not help. A singular ResourseHubLinks feature tests is 2:30 and prevents simple gains from simple parallelization increase.

Issue #2061, opened 2025-03-05
Automatically populate .env file with login related env vars


shiroyasha:
Login related env vars that should be set to true are:

```
ALLOW_LOGIN_WITH_EMAIL=yes
ALLOW_SIGNUP_WITH_EMAIL=yes
```

We can use the same approach as here to add the env vars if they are not already present:
https://github.com/operately/operately/blob/main/Makefile#L80

caiquefortunato:
I made the suggested change and created the Pull Request :)

PR #2226, opened 2025-04-03
Ensure email login/signup flags are present in .env during setup

caiquefortunato:
This PR resolves Issue #2061: ensures that the following environment variables are present in the .env file when running make dev.setup:
	•	ALLOW_LOGIN_WITH_EMAIL=yes
	•	ALLOW_SIGNUP_WITH_EMAIL=yes

These variables are required to enable email-based login and signup in development.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,6 +82,8 @@ dev.run.script:
 dev.seed.env:
 	touch .env
 	grep "OPERATELY_BLOB_TOKEN_SECRET_KEY" .env || echo "OPERATELY_BLOB_TOKEN_SECRET_KEY=$$(openssl rand -base64 32)" >> .env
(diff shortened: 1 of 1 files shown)


caiquefortunato:
I used `printf` to ensure that the content added includes an explicit newline (`\n`) before it, preventing the new text from being appended directly to the end of the last line—especially if the original file doesn't end with a newline. `printf` gives me precise control over special characters, which `echo` doesn't guarantee in the same way.

shiroyasha:
Looks great! Thanks for the contribution.

We've done a [larger code transition](https://discord.com/channels/1080898715268698152/1081189157533134919/1356998471676198993) of the repository this week, so I'll wait for that work to settle before I merge this PR. My estimate is that I can come back to merging this later during the day, or on monday.

PR #2251, opened 2025-04-08
chore: Extract Statusbadge component to turboui

markoa:
- Moved the component from `design/` to `turboui/`
- Created a showcase page in design system
- I don't consider the color and state combinations to be final, but changing them will be trivial
CleanShot 2025-04-08 at 14 34 43@2x


markoa:
I've fixed the Tailwind issue @shiroyasha.

PR #2285, opened 2025-04-14
chore: Extract ResourceName component to turboui

Rockyy174:
I've extracted ResourceName component to turboui.


shiroyasha:
ResourceName doesn't sound like a generic component that we can use in multiple places. 
What is it?

Rockyy174:
It renders the  `IconTarget + GoalName` or  `IconHexagons + ProjectName`. 

Do you think I should choose another name for this component? Or do you think this component isn't reusable enough to be in the `turboui`?

shiroyasha:
The name is too generic, resource can mean anything. Lets keep it as part of a `WorkMap` component. Like in the mini-work-map: https://storybooks-aq1.pages.dev/?path=/docs/components-miniworkmap--docs.

PR #2379, opened 2025-05-05
fix: Use dev JS loading strategy for local tests to fix Wallaby failures

Rockyy174:
Fixes Wallaby feature tests that were failing with MIME type errors when loading JavaScript modules. The solution uses the Vite dev server for local tests while maintaining the production asset loading for CI environments.

### DIFF
--- a/app/lib/operately_web/controllers/page.html.heex
+++ b/app/lib/operately_web/controllers/page.html.heex
@@ -46,7 +46,7 @@
       2. Production strategy: Used in production or in CI test environments.
          JavaScript is loaded from the static assets that are built by Vite.
     -->
-    <%= if Application.get_env(:operately, :app_env) == :dev or (Application.get_env(:operately, :app_env) == :test and System.get_env("CI") == nil) do %>
+    <%= if OperatelyWeb.PageController.development_mode?() do %>
       <script type="module">
         import RefreshRuntime from 'http://localhost:4005/@react-refresh'
         RefreshRuntime.injectIntoGlobalHook(window)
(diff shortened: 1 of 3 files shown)


app/lib/operately_web/controllers/page.html.heex
```diff
@@ -37,15 +37,16 @@
...
-      assets that are built by Vite.
+      1. Development strategy: Used in development mode or in local test mode (non-CI).
+         JavaScript is loaded from the local Vite server, which allows for hot reloading
+         and other development features.
+
+      2. Production strategy: Used in production or in CI test environments.
+         JavaScript is loaded from the static assets that are built by Vite.
     -->
-    <%= if Application.get_env(:operately, :app_env) == :dev do %>
+    <%= if Application.get_env(:operately, :app_env) == :dev or (Application.get_env(:operately, :app_env) == :test and System.get_env("CI") == nil) do %>
```
shiroyasha, line 49: Maybe we could extract this login into the controller (page.html.ex).

Rockyy174: Sure, makes sense. The thing is that it isn't working in Semaphore. I tried both `System.get_env("CI") == nil` and `System.get_env("CI") != "true"`. 

Do you know what could be the problem?

shiroyasha: Yes. The CI env is available in the bash environment, but when we run the tests with docker-compose the variable is not passed into the docker environment.

Maybe `echo CI=true >> .env` in the semaphore.yml would solve the situation? Or maybe even better, in the test.seed.env makefile target.

Rockyy174: I will give it a try. Thanks! 🙇

PR #2389, opened 2025-05-06
fix: Don't reload WorkMap data when tab changes

Rockyy174:
Now, when Work Map tab changes, the Work Map data isn't reloaded.


app/assets/js/pages/index.tsx
```diff
@@ -183,6 +183,7 @@ export default {
     name: "CompanyWorkMapPage",
     loader: CompanyWorkMapPage.loader,
     Page: CompanyWorkMapPage.Page,
```
shiroyasha, line 185: This file is recreated from scratch when we run `make gen`, right?

Rockyy174: I just checked it. It turned out it is 🤔

Rockyy174:
@shiroyasha What do you think about this solution? Now every route includes `shouldRevalidate`, which can be defined on per-page basis.

shiroyasha:
Looks good. :shipit:

PR #2534, opened 2025-05-23
chore: Add Work Map behavior

Rockyy174:
I've added a behavior for the Work Map Item. It requires the following functions from resources that are implementing this behavior:

- status
- state
- next_step
- progress_percentage

### DIFF
--- a/app/lib/operately/goals/goal.ex
+++ b/app/lib/operately/goals/goal.ex
@@ -2,9 +2,9 @@ defmodule Operately.Goals.Goal do
   use Operately.Schema
   use Operately.Repo.Getter
 
+  alias Operately.WorkMaps.WorkMapItem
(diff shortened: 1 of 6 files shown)


app/lib/operately/goals/goal.ex
```diff
@@ -82,14 +82,91 @@ defmodule Operately.Goals.Goal do
...
+    target_progresses = Enum.map(targets, &target_progress_percentage/1)
+
+    if Enum.empty?(target_progresses) do
+      0
+    else
+      Enum.sum(target_progresses) / length(target_progresses)
+    end
+  end
+
+  defp target_progress_percentage(target) do
```
shiroyasha, line 149: This (Goal module) is not the right module for this. Move it to Goals.Target.

Issue #2518, opened 2025-05-22
My work


markoa:
My work will be a core page in top navigation one of the last pieces of the work map puzzle.

As a reminder we've been planning to change the (first half of) top navigation to
- Home
- Company (shows work map)
- My work
- Review

Side note: since Operately is tracking multiple types of work (goals, projects, tasks, and more in the future — but also work I'm responsible to review) then it's not yet 100% clear what the best version of this screen is. But we are not solving that problem right now. We're solving the problem of making the data visible and available to begin with.

So I propose this:

- Change the layout of personal profile pages to be full width and use the new turboui tabs.
	- The idea is to immediately make it so what you see about yourself is what others see about yourself too (filtered by any permission requirements ofc). Two birds one stone.
- Change tabs to
	- Assigned
	- Reviewing
	- Recent activity (move the feed there)
	- About (stays as is)

In the Assigned tab, we show a variation of work map with links to toggle between the following (see Linear's 'My issues' screen for an example):
- Overview: shows flat lists of active goals and projects one below another; filtering will not be possible here; no need to show 'Champion' column ofc
- Goals (flat list)
- Projects
- Completed

Same approach in Reviewing tab, except Completed doesn't apply there.

PR #2571, opened 2025-05-29
chore: GetWorkMapQuery now supports champion_id, reviewer_id and contributor_id as filters

Rockyy174:
I'm extending the `GetWorkMapQuery` so that it also supports filtering by `champion_id`, `reviewer_id` and `contributor_id`.

These changes will be necessary to implement the "My work" page as described in https://github.com/operately/operately/issues/2518.


app/lib/operately/work_maps/get_work_map_query.ex
```diff
@@ -18,15 +18,16 @@ defmodule Operately.WorkMaps.GetWorkMapQuery do
...
+    - contributor_id (optional): The ID of a contributor
     - include_assignees (optional): A boolean indicating whether to include assignees in the result. Defaults to false
   """
   def execute(person, args) do
     company_id = Map.get(args, :company_id)
-    include_assignees = Map.get(args, :include_assignees, false)
 
-    goals_task = Task.async(fn -> get_goals(person, company_id, include_assignees) end)
-    projects_task = Task.async(fn -> get_projects(person, company_id, include_assignees) end)
+    goals_task = Task.async(fn -> get_goals(person, company_id, args) end)
```
shiroyasha, line 29: It seems that we could leave the company_id in the args to have less parameters to shuffle around.

Issue #2572, opened 2025-05-30
Standardize nullable, optional, and required flags in the api


shiroyasha:
- [x] 1. Remove required: true
- [x] 2. Rename nullable to null
- [x] 3. Introduce field? generator
- [x] 4. Switch the default from null: true, null: false
- [x] 5. Introduce default: val

PR #2694, opened 2025-06-14
chore: Change api default from null: true to null: false

shiroyasha:


shiroyasha:
fixes https://github.com/operately/operately/issues/2572

PR #3449, opened 2025-09-17
feat: Throw confetti on milestone page when user clicks on 'Mark complete'

markoa:
because completing a milestone is a big deal

### DIFF
--- a/turboui/src/utils/confetti.ts
+++ b/turboui/src/utils/confetti.ts
@@ -1,0 +1,29 @@
+type ConfettiModule = typeof import("canvas-confetti");
+type ConfettiFn = ConfettiModule["default"];
+type ConfettiOptions = Parameters<ConfettiFn>[0];
+
+let loader: Promise<ConfettiFn> | null = null;
+
+export function launchConfetti(options?: ConfettiOptions) {
+  if (typeof window === "undefined") return;
+
+  if (!loader) {
+    loader = import("canvas-confetti").then((module) => module.default);
+  }
+
+  loader
+    .then((confetti) => {
+      confetti({
+        spread: 70,
+        particleCount: 240,
+        origin: { y: 0.6 },
+        disableForReducedMotion: true,
+        ...options,
+      });
+    })
+    .catch((error) => {
+      if (process.env.NODE_ENV !== "production") {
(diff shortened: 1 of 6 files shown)


turboui/src/utils/confetti.ts
```diff
@@ -0,0 +1,29 @@
...
+type ConfettiFn = ConfettiModule["default"];
+type ConfettiOptions = Parameters<ConfettiFn>[0];
+
+let loader: Promise<ConfettiFn> | null = null;
+
+export function launchConfetti(options?: ConfettiOptions) {
+  if (typeof window === "undefined") return;
+
+  if (!loader) {
+    loader = import("canvas-confetti").then((module) => module.default);
```
shiroyasha: This is super suspicious. Why do we need to dynamically load this?

turboui/src/utils/confetti.ts
```diff
@@ -0,0 +1,29 @@
...
+      confetti({
+        spread: 70,
+        particleCount: 240,
+        origin: { y: 0.6 },
+        disableForReducedMotion: true,
+        ...options,
+      });
+    })
+    .catch((error) => {
+      if (process.env.NODE_ENV !== "production") {
```
shiroyasha: We are never going to run this in Node. No need for this kind of checks.

turboui/src/utils/confetti.ts
markoa: "canvas-confetti is browser-only: it pokes at the DOM and the window object during module evaluation. In our setup the UI library is bundled for SSR/Storybook and also shared with the Phoenix app, so eagerly importing it at the top level would:
- blow up the server-side render (no window), and
- ship the whole library into every bundle even though we use it in one narrow path.

The dynamic import keeps it client-only and defers the cost until someone actually marks a milestone complete. Once loaded, we memoize the resolved default export so subsequent launches reuse the same function with no extra fetch."

shiroyasha: I assume this is codex.

1/ Afaik, we are never doing SSR, not even in storybooks.
2/ We are always bundling the whole app into one bundle, no need for this optimization that can force us to change how we deploy our assets.

Could be that I'm missing something 🤷‍♂️

PR #3562, opened 2025-10-01
feat: Invite link landing page

shiroyasha:


shiroyasha:
Good point. It was inertia from the state where you have two competing options.

PR #4057, opened 2026-01-19
fix: autosave rich text drafts

Rockyy174:
### Motivation
- Prevent accidental loss of user-typed rich text when a page is reloaded, navigated away from, or otherwise interrupted while keeping drafts ephemeral rather than persisted long-term.

### Description
- Add session-based autosave/restore to `RichTextArea` by importing `isContentEmpty` from `turboui` and adding an optional `autosaveKey` prop. 
- Compute a default `autosaveKey` using `window.location.pathname`, `window.location.search`, and the field name when one is not provided.
- On editor updates store `{ savedAt, content }` into `sessionStorage` unless content is empty, and remove the stored draft when content is empty.
- On mount, if the current value is empty try to restore a draft from `sessionStorage` only if it is present and not older than a short TTL (`AUTOSAVE_TTL_MS`, 6 hours), otherwise clear stale data.

### Testing
- No automated tests were run for this change.
- Manual runtime safeguards are present: operations that read/write `sessionStorage` are wrapped in `try/catch` and failures are logged without breaking the editor.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_b_696e0d6c9644832ab221a8c959c91793)

Part of https://github.com/operately/operately/issues/1576.

### DIFF
--- a/app/assets/js/components/Forms/RichTextArea.tsx
+++ b/app/assets/js/components/Forms/RichTextArea.tsx
@@ -7,7 +7,7 @@ import { useValidation } from "./validations/hook";
 
 import classNames from "classnames";
 import { validateRichContentPresence } from "./validations/richContentPresence";
-import { isContentEmpty, useEditor, Editor as RichTextEditor, RichContent } from "turboui";
+import { useEditor, Editor as RichTextEditor, RichContent } from "turboui";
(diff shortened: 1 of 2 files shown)


markoa:
@Rockyy174 this functionality is quite needed - it's really easy to accidentally refresh / close the input box. I will open an enhancement issue.

PR #4362, opened 2026-04-15
fix: improve generated API docs metadata

markoa:
Updated the generator in `app/lib/operately/api_docs/markdown.ex:8` so human-facing API docs text is derived centrally instead of using raw slugs or generic boilerplate. Tests were updated in `app/test/operately/api_docs/markdown_test.exs:6` and `app/test/operately/api_docs/generator_test.exs:16`.

  Titles and descriptions now come from a small helper set in `markdown.ex`:

  - Namespace title: companies -> Companies API, resource_hubs -> Resource Hubs API
  - Endpoint title: <Namespace> API: <Humanized operation> such as Companies API: Create member
  - Endpoint description: <Humanized operation> with the Operately <Namespace> API.
  - Visible labels in namespace tables and root index are humanized from slugs, while routes and filenames
    stay unchanged

### DIFF
--- a/app/lib/operately/api_docs/markdown.ex
+++ b/app/lib/operately/api_docs/markdown.ex
@@ -12,7 +12,7 @@ defmodule Operately.ApiDocs.Markdown do
       |> Enum.map(fn endpoint ->
         """
         <tr>
-          <td><a href="./#{endpoint.name}">#{endpoint.name}</a></td>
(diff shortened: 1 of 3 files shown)


Rockyy174 (approved):

Rockyy174:
@markoa I pulled and tested it. Everything works!

PR #4571, opened 2026-05-27
fix: Correctly verify Polar billing webhooks

Rockyy174:


app/lib/operately/billing/polar/webhook_verifier.ex
```diff
@@ -74,11 +79,56 @@ defmodule Operately.Billing.Polar.WebhookVerifier do
...
+      end)
+  end
+
+  defp candidate_secret_keys(secret) when is_binary(secret) do
+    # Polar follows Standard Webhooks; some docs/examples describe secrets as base64-encoded key
+    # material, but real-world setups sometimes behave as if the literal secret string is the key.
+    #
+    # To avoid false negatives across environments, we try both:
+    # - the decoded suffix (when `polar_whs_...` looks like base64/base64url)
+    # - the literal secret string bytes
```
shiroyasha: Do they really send two different kinds?

Rockyy174: Likely not, but I didn't see any harm in keeping support for both. That was the fix for the mismatch I had mentioned.

Rockyy174: But now that you pointed it out, I think it's cleaner and probably safe enough to keep support for only one kind.

PR #4629, opened 2026-06-04
feat: add docs & files to projects

### LINKED ISSUE #813: Docs & Files in projects

mancinux:
resolves #813.


app/assets/js/pages/ProjectPage/index.tsx
```diff
@@ -335,10 +383,15 @@ function Page() {
...
 
-    resources,
-    onResourceAdd: createResource,
-    onResourceEdit: updateResource,
-    onResourceRemove: removeResource,
+    docsAndFiles: resourceHub
+      ? {
+          preview: <DocsAndFilesPreview resourceHub={resourceHub} nodes={resourceHubNodes} />,
+          tabContent: (
+            <DocsAndFilesTab resourceHub={resourceHub} nodes={resourceHubNodes} draftNodes={resourceHubDraftNodes} />
```
Rockyy174: I think this might get messy. These new pages defined in turboui are supposed to handle all the UI while the app handles the interactions with the backend. 

I saw that we even had to mock these components in storybooks, which probably means storybooks doesn't show the docs tab 100% accurately and will just get worse over time.

I can see that lots of components used in DocsAndFilesPreview and DocsAndFilesTab are defined in the app, because they were developed before turboui was introduced. 

My suggestion is that we move DocsAndFilesPreview, DocsAndFilesTab and the components they are using to turboui. The other Resource Hub pages that also rely on these components can import from turboui as well.

app/assets/js/pages/ProjectPage/index.tsx
```diff
@@ -676,78 +716,109 @@ function useMilestones(paths: Paths, project: Projects.Project, refresh?: () =>
...
+function DocsAndFilesTab({
+  resourceHub,
+  nodes,
+  draftNodes,
+}: {
+  resourceHub: ResourceHubs.ResourceHub;
+  nodes: ResourceHubs.ResourceHubNode[];
+  draftNodes: ResourceHubs.ResourceHubNode[];
+}) {
+  const { refresh } = PageCache.useData(loader, { refreshCache: false });
```
Rockyy174: It seems that when you add a folder or file, they only appear after the page is refreshed.

app/assets/js/pages/ProjectPage/index.tsx
mancinux: thanks for the feedback, addressed

Rockyy174: That's not what I meant. Now we have the same components defined in both app for the Project page and in turboui for storybooks, which is not correct. The components should be defined in turboui only. If app needs those components, it can import them from turboui. 

However, ProjectPage is already defined in turboui, turboui/src/ProjectPage/index.tsx, so these components should be used directly on the page component. They don't need to be imported by app and then passed as props to ProjectPage.

app/lib/operately/resource_hubs/project_hub.ex
```diff
@@ -0,0 +1,73 @@
...
+  @name "Documents & Files"
+
+  def create_for_project(project = %Project{}) do
+    with {:ok, hub} <- find_or_create_hub(project),
+         {:ok, _context} <- ensure_context(hub),
+         {:ok, _} <- sync_access_from_project(project.id) do
+      {:ok, hub}
+    end
+  end
+
```
Rockyy174, line 17: Are we using `ProjectHub.sync_access_from_project/1` and `SpaceHub.sync_access_from_space/1` to ensure the users always have the same access to resource hub as they have to the parent space or project?

If yes, I don't think syncing is the best approach. It's easy to forget about it in some operations and it can easily add about a dozen calls to the db in a single operation.

When child resources rely on the same permissions as the parent, we just use the same access_context as the parent for the children. 

For most resources, that can be achieved very easily by adding `has_one :access_context, through: :parent`.

For polymorphic resources like this one, it's a bit more tricky, but you can achieve the same result by creating a wrapper around `Getter` and adjusting the base query, like it was done in `Operately.Tasks.Task`.

PR #4644, opened 2026-06-09
chore: Add skills for writing docs

Rockyy174:


markoa:
This isn't quite what I had in mind. Instructions on how to write would be best to have in the website repo, so any agent working on the help section picks it up.

Here we're interested in a skill that goes through app's git history after [sha/tag given as argument] and builds a list of things to document, inferring from code how things works.

markoa:
So we have a pipeline of specialized agents
- agent1 parses code changes and prepares a list of changes in probably a bit terse/ technical language
- agent2, based on writing guidelines in website repo, converts that into nice content that non technical users and search engines understand

Rockyy174:
@markoa I think now it meets what you described. 

I tested the skills by asking an agent to list what needs documentation since commit 52814755b (random commit). This was the result: https://gist.github.com/Rockyy174/dca9093c6ba7dc244b4998ce050896e9. It looks accurate.

PR #5016, opened 2026-08-06
fix: Remove search extra row from Docs & Files tab

### LINKED ISSUE #5007: Change copy and move 'Search this resource hub' piece
Image

No need to introduce a whole new row for it. Plus we don't use the term resource hub in UI. Just say "Search" instead.

Rockyy174:
Fixes https://github.com/operately/operately/issues/5007

### DIFF
--- a/turboui/src/DocsAndFiles/PageDocsAndFiles.tsx
+++ b/turboui/src/DocsAndFiles/PageDocsAndFiles.tsx
@@ -8,6 +8,7 @@ import {
   NewFileModalsProvider,
   NodeMenu,
   ResourceHubNodesListProvider,
+  ResourceHubSearchInput,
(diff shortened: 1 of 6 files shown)


Rockyy174:
I went back in the git history and it turned out that the extra row was not introduced by the search box. It was already there because of the items count and the filters button. 

To remove the extra row, I decided to remove the counter and move the filters to the above row too. We now have the counter in the tabs, so we didn't lose any information. 

To the left, besides the title, I don't think it looks good:
CleanShot 2026-08-05 at 21 35 37

To the right, I think it looks okay:
CleanShot 2026-08-05 at 21 41 57

@markoa Let me know if you like this layout. If you think it's good, it's ready for merging.

PR #5031, opened 2026-08-08
fix: Render API-created space task notes

### LINKED ISSUE #4976: Space task description (Notes) still not rendered in TaskSlideIn after upgrading to 1.7.0 (related to #4518 / #4519)
## Title
Space task description (Notes) still not rendered in TaskSlideIn after upgrading to 1.7.0 (related to #4518 / #4519)

## Summary

After upgrading our self-hosted instance to **1.7.0**, which includes the fix in #4519 for #4518 ("Task description set via API not visible in UI (space tasks)"), the description ("Notes" in the UI) of a Space Task created via the external API is still not rendered in the TaskSlideIn panel, even though the description is correctly persisted and returned by the API.

This reproduces both:
- when `description` is sent directly in the `tasks/create` payload, and
- when `description` is set afterwards via a separate `tasks/update_description` call

## Environment

- Operately version: 1.7.0 (self-hosted, Docker Compose / Dokploy)
- Task type: Space task (not Project task)
- Browser tested: Arc, including a clean incognito session (no prior cache/session)
- Cloudflare edge cache purged for the instance domain before testing

## Steps to reproduce

1. Create a Space task via the external API `tasks/create` endpoint, including a `description` field directly in the payload (ProseMirror doc, JSON-stringified), e.g.:
   ```json
   {
     "type": "space",
     "id": "<space_id>",
     "name": "Test task",
     "assignee_id": null,
     "description": "{\"type\":\"doc\",\"content\":[{\"type\":\"paragraph\",\"content\":[{\"type\":\"text\",\"text\":\"Hello world\"}]}]}",
     "due_date": { "date": "2026-08-10", "date_type": "day", "value": "Aug 10, 2026" },
     "status": { "id": "<pending_status_id>", "value": "pending", "label": "Pending", "color": "gray", "index": 0, "closed": false }
   }
   ```
2. Confirm via `GET /api/v2/spaces/list_tasks?space_id=<space_id>` that the created task's `description` field is populated correctly and matches what was sent. (Confirmed in our case — see "API response" section below.)
3. In a clean browser session (incognito, no cache), navigate directly to the Space page and open the task's TaskSlideIn panel.
4. Observe the "Notes" section — it renders as empty, with only an "Edit" button, despite the API returning a non-empty `description`.

## Expected behavior

The "Notes" panel should render the rich text content from the `description` field, matching what's returned by `spaces/list_tasks`.

## Actual behavior

"Notes" panel is empty. No errors are shown in the UI.

## Additional diagnostics performed

- Verified via Network tab that opening the TaskSlideIn panel does **not** trigger any additional XHR call to fetch task details (e.g. no `tasks/get` call) — the panel appears to rely entirely on data already available client-side (from `spaces/list_tasks` or a realtime subscription), which we confirmed does contain the correct `description`.
- Ruled out CDN/edge caching (Cloudflare cache purged) and browser caching (tested in a clean incognito session with no prior visits to the instance).
- Ruled out backend persistence issues — the same `description` field, when queried directly via `spaces/list_tasks`, is present and correctly formed for both:
  - tasks created via `tasks/create` with `description` included directly, and
  - tasks created via `tasks/create` (no description) followed by a separate `tasks/update_description` call.
- This suggests the issue is isolated to how `PageDescription` (or its parent component in TaskSlideIn) receives/parses `description` on initial mount — possibly independent of the async-arrival scenario that #4519 fixed, since in our case the data is already present at mount time.

## API response (for a task with the issue)

```json
{
  "id": "dc-2026-0023-osobn-odb-r-GH4QJangngBOQBxn3QQUmU",
  "name": "#DC-2026-0023 · Osobní odběr Praha · Matěj Malecha",
  "type": "space",
  "description": "{\"content\":[{\"content\":[{\"text\":\"Zákazník: Name Surname | email@test.com | \",\"type\":\"text\"}],\"type\":\"paragraph\"},{\"content\":[{\"text\":\"\",\"type\":\"text\"}],\"type\":\"paragraph\"},{\"content\":[{\"text\":\"Doprava: Osobní odběr Praha\",\"type\":\"text\"}],\"type\":\"paragraph\"},{\"content\":[{\"text\":\"\",\"type\":\"text\"}],\"type\":\"paragraph\"},{\"content\":[{\"text\":\"Cena celkem: 250 Kč\",\"type\":\"text\"}],\"type\":\"paragraph\"},{\"content\":[{\"text\":\"\",\"type\":\"text\"}],\"type\":\"paragraph\"},{\"content\":[{\"text\":\"Položky:\",\"type\":\"text\"}],\"type\":\"paragraph\"},{\"content\":[{\"text\":\"– Standardni vydani × 1\",\"type\":\"text\"}],\"type\":\"paragraph\"}],\"type\":\"doc\"}",
  "inserted_at": "2026-08-03T13:50:23Z",
  "updated_at": "2026-08-03T13:50:23Z",
  "status": {
    "closed": false,
    "id": "38fed8f4-248d-4993-8070-15ec246bfef5",
    "index": 0,
    "label": "Čeká na vyřízení",
    "value": "pending",
    "color": "gray"
  }
}
```

Note that `inserted_at` equals `updated_at`, confirming the description was set at creation time and never subsequently modified — the data has been correct since the task's creation.

## Notes

We're aware of #4518 / #4519 which addressed a similar-sounding symptom, but the fix there specifically targeted the case where `description` arrives *after* the component has already mounted (`PageDescription` getting stuck in "zero" mode). In our reproduction, the description is already present in the API response at the very first client-side fetch, which suggests this may be a distinct issue rather than a regression of the original one — though it's possible both share a root cause in how `PageDescription` initializes state.

Happy to provide further diagnostics (HAR file, additional API responses, etc.) if useful.

mvanhorn:
Normalize rich-text documents at the shared `useEditor` boundary before both initial TipTap construction and later `setContent` updates, removing zero-length text nodes while preserving meaningful text, marks, mentions, and empty paragraph structure. Space tasks created or updated through the external API can persist a non-empty ProseMirror description that appears blank in the task slide-in.

A `PageDescription` whose document mixes meaningful paragraphs with zero-length text nodes renders all meaningful text instead of an empty editor; A valid rich-text document remains unchanged by normalization, while empty paragraphs remain valid empty blocks and non-text nodes such as mentions retain their attributes and marks.

Fixes #4976

### DIFF
--- a/app/test/features/space_kanban/task_cards_test.exs
+++ b/app/test/features/space_kanban/task_cards_test.exs
@@ -38,18 +38,32 @@ defmodule Operately.Features.SpaceKanban.TaskCardsTest do
     |> Steps.assert_description(content: "Updated description for kanban flow.")
   end
 
-  feature "task description set through external API is visible in the slide-in", ctx do
-    description = Operately.Support.RichText.rich_text("Hello world from API", :as_string)
+  feature "task description set during external API creation is visible in the slide-in", ctx do
+    ctx =
+      ctx
+      |> Factory.add_api_token(:api_token, :creator, read_only: false)
+      |> Steps.create_space_task_through_external_api(
+        name: "External API task",
+        description: api_task_description()
+      )
 
+    ctx
+    |> Steps.visit_kanban_page(task_id: ctx.external_api_task_id)
+    |> Steps.assert_description(content: "First paragraph from the API.")
+    |> Steps.assert_description(content: "Second paragraph from the API.")
+  end
+
+  feature "task description updated through external API is visible after a fresh page load", ctx do
     ctx =
       ctx
       |> Factory.add_api_token(:api_token, :creator, read_only: false)
       |> Steps.create_space_task_through_external_api(name: "External API task")
-      |> Steps.update_task_description_through_external_api(description: description)
+      |> Steps.update_task_description_through_external_api(description: api_task_description())
 
     ctx
     |> Steps.visit_kanban_page(task_id: ctx.external_api_task_id)
-    |> Steps.assert_description(content: "Hello world from API")
+    |> Steps.assert_description(content: "First paragraph from the API.")
+    |> Steps.assert_description(content: "Second paragraph from the API.")
   end
 
   feature "delete a task from the slide-in", ctx do
@@ -62,5 +76,27 @@ defmodule Operately.Features.SpaceKanban.TaskCardsTest do
     |> Steps.change_task_status(prev_status: "Not started", next_status: new_status)
     |> Steps.delete_task()
     |> Steps.assert_task_removed(task_key: :second_task, status_value: new_status)
+  end
+
+  defp api_task_description do
+    Jason.encode!(%{
+      type: "doc",
+      content: [
+        %{
+          type: "paragraph",
+          content: [
+            %{type: "text", text: ""},
+            %{type: "text", text: "First paragraph from the API."}
+          ]
+        },
+        %{
+          type: "paragraph",
+          content: [
+            %{type: "text", text: "Second paragraph from the API."},
+            %{type: "text", text: ""}
+          ]
+        }
+      ]
+    })
   end
 end
--- a/app/test/support/features/space_kanban_steps.ex
+++ b/app/test/support/features/space_kanban_steps.ex
@@ -42,16 +42,25 @@ defmodule Operately.Support.Features.SpaceKanbanSteps do
     |> UI.assert_has(testid: "task-slide-in")
   end
 
-  step :create_space_task_through_external_api, ctx, name: name do
+  step :create_space_task_through_external_api, ctx, opts do
     conn = api_conn()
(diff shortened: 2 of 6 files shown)


mvanhorn:
Done, and it was worth doing. Pulled the normalization out into `richTextContent.ts` with typed `RichTextNode` and `RichTextContent` definitions, so the helper no longer takes `any` and misuse is caught at compile time. Moved the focused coverage alongside it in `richTextContent.test.ts` and updated `useEditor.tsx` to the new signature.

Rockyy174:
@mvanhorn Thanks for solving this issue. The changes look good. Could you please sign off the commits?

mvanhorn:
@Rockyy174 done, both commits are signed off now.

While amending I also rewrote the first commit message: it had absolute paths from my local checkout in it, which were meaningless to anyone else reading the history. The message now refers to the files by repository-relative path. The tree is byte-identical to what you reviewed, so only the messages changed.

PR #5129, opened 2026-08-25
fix: Plot the same KPI history on the space card and the KPI list

shiroyasha:
## Summary

The space dashboard KPI card and the KPIs tool list render the same `KpiSparkline` component, but they were fed different data:

- `kpis/list_kpis` (the KPI list) loads a bounded window of the **12 most recent** entries per KPI via `Kpis.load_recent_entries/2`, and sets the virtual `latest_entry` field.
- `spaces/list_tools` (the space dashboard card) preloaded the **entire** entry history and never set `latest_entry`.

The same KPI could therefore show a different line shape, a different trend colour (the sparkline colours by last-vs-first plotted point), and even a different latest value on the two screens — older history included on one side, excluded on the other.

This extracts the list-loading into `Kpis.list_kpis_with_recent_entries/1` and uses it from both endpoints, so every place that lists KPIs plots the same series and reports the same latest entry.

Loading the whole history for every KPI in a space was also unnecessary work for a 56px-wide sparkline, so this is a small win on the space dashboard query too.

### Before

The same three KPIs, same moment. "GitHub stars" trends green in the list (recent 12 periods) and red on the card (full history), with different latest values.

| KPIs tool list | Space dashboard card |
| --- | --- |
|  |  |

## Changes

- `Operately.Kpis.list_kpis_with_recent_entries/1` — one loader for "KPIs of a space, ready to list": champion preloaded plus the bounded recent-history window.
- `OperatelyWeb.Api.Kpis.ListKpis` and `OperatelyWeb.Api.Spaces.ListTools` both call it, removing the duplicated (and divergent) query in `list_tools`.
- Factory: `enable_space_tool/disable_space_tool` (and `enable_tool/disable_tool`) now support `:kpis`, and `set_tools` passes `:kpis_enabled` through. The KPIs tool was the only one missing from these helpers, so it could not be toggled in tests.

## Test plan

- [x] `make test FILE=app/test/operately/kpis/kpis_test.exs` — covers the new loader (scoping, champion preload, 12-entry cap, `latest_entry`).
- [x] `make test FILE=app/test/operately_web/api/spaces/list_tools_test.exs` — new `kpis` describe block asserting the endpoint returns the same 12-entry window and `latest_entry` as the KPIs tool, plus the empty-history and tool-disabled cases.
- [x] `make test FILE=app/test/operately_web/api/kpis/list_kpis_test.exs` — unchanged behaviour after the refactor.

No migration and no API type changes: `entries` and `latest_entry` were already part of the serialized `:kpi` object, `list_tools` simply was not populating `latest_entry`.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/app/lib/operately/kpis.ex
+++ b/app/lib/operately/kpis.ex
@@ -18,6 +18,17 @@ defmodule Operately.Kpis do
     |> Repo.all()
   end
 
+  @doc """
+  Lists a space's KPIs ready to be rendered in a list: champion preloaded, and a
+  bounded window of recent history per KPI. Every place that lists KPIs uses this
+  so they all plot the same series.
+  """
+  def list_kpis_with_recent_entries(space_id) do
+    list_kpis(space_id)
+    |> Repo.preload(:champion)
+    |> load_recent_entries()
+  end
+
   def get_kpi(id), do: Repo.get(Kpi, id)
   def get_kpi!(id), do: Repo.get!(Kpi, id)
 
--- a/app/lib/operately/kpis.ex
+++ b/app/lib/operately/kpis.ex
@@ -37,7 +37,7 @@ defmodule Operately.Kpis do
     |> Repo.all()
   end
 
-  # How much history the list view carries per KPI: enough to plot a trend
+  # How much history a list view carries per KPI: enough to plot a trend
   # inline, without the cost of loading the whole series for every KPI.
   @recent_entries_limit 12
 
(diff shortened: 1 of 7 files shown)


shiroyasha:
Good catch on the stale comment — updated in d52684a6f. The `latest_entry` doc now points at `Kpis.load_recent_entries/2` rather than `ListKpis`, and the `@recent_entries_limit` comment reads "a list view" since both the KPIs tool and the space dashboard card now share it.

PR #5269, opened 2026-09-11
chore: Add feed performance benchmark

Rockyy174:


Rockyy174:
@shiroyasha This is the benchmark I used to debug the feed and validate the solution, but I'm not sure if it would be valuable to have it in the codebase. Do you think I should merge or discard it?

shiroyasha:
We can discard.

